PR #123, opened 2016-11-21
Changing function StorageUploadFile to accept a readseeker in place of go-openapi runtime.File

ederprado-luizalabs:


alexguerra-luizalabs:
:+1:

PR #127, opened 2016-11-28
Fixes for the apps logs endpoint

alexguerra-luizalabs:
- only consider running pods
- avoid a race condition when accessing a pointer to a range variable inside
  a goroutine; this was causing the selection of the wrong pods
- simplify code by avoiding bufio
- don't leak fds

### DIFF
--- a/k8s/apps.go
+++ b/k8s/apps.go
@@ -1,7 +1,6 @@
 package k8s
 
 import (
-	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -46,6 +45,7 @@ type apps struct {
 type podListLogger struct {
 	podList *api.PodList
 	r       io.ReadCloser
+	w       io.WriteCloser
 }
 
 func newApps(c *k8sHelper) *apps {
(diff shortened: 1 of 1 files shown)


k8s/apps.go
```diff
@@ -549,24 +549,22 @@ func streamPodOutput(k *k8sHelper, pod *api.Pod, opts *api.PodLogOptions) (strea
 
 func (c apps) newPodListLogger(podList *api.PodList, opts *api.PodLogOptions) io.ReadCloser {
 	var wg sync.WaitGroup
-	wg.Add(len(podList.Items))
 	r, w := io.Pipe()
 	for _, pod := range podList.Items {
-		p := &pod
+		tmp := pod
+		if tmp.Status.Phase != api.PodRunning {
+			continue
```
arnaldopereira: ❤️

k8s/apps.go
```diff
@@ -549,24 +549,22 @@ func streamPodOutput(k *k8sHelper, pod *api.Pod, opts *api.PodLogOptions) (strea
...
+		}
+		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			rc, err := streamPodOutput(c.k, p, opts)
+			rc, err := streamPodOutput(c.k, &tmp, opts)
 			if err != nil {
 				log.Errorf("reading logs %s", err)
```
arnaldopereira: I'd take the chance you're touching this code to improve this log message: `reading logs` doesn't help much.

PR #153, opened 2017-02-08
Set revision history limit per app

drgarcia1986:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/153)


drgarcia1986:
[Revision History Limit](https://kubernetes.io/docs/user-guide/deployments/#revision-history-limit) is the limit of revisions (i.e. deploys) that kubernetes need to store, the default value is without limit and for this reason each deploy store a revision and use resources of the k8s cluster.

PR #168, opened 2017-03-20
Add support for non-web process types

### LINKED ISSUE #142: Support non web process types
Today only the web process is started. Suggestion: multi-container pods.

alexguerra-luizalabs:
Following Heroku's tradition of not automatically running dynos or
routing web traffic to non-web processes and independent scaling,
we associate to each app one process type and don't create
kubernetes services when this process is non-web. Fixes #142.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/168)


k8s/apps.go
```diff
@@ -86,7 +92,25 @@ func (c apps) Create(app *models.App, storage helpers.Storage, tk *Token) error
 	if err := c.k.Deployments().CreateAutoScale(app); err != nil {
 		return err
 	}
-	// creating the loadbalancer
+	// Creating the service
+	if err := c.createOrSkipService(app); err != nil {
+		return err
+	}
+	return nil
```
drgarcia1986: Why not?
```go
return c.createOrSkipServevice(app)
```
?

models/app_in.go
```diff
@@ -36,6 +36,12 @@ type AppIn struct {
 	*/
 	Name *string `json:"name"`
 
+	/* app process type
+
+	Pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	*/
+	ProcessType *string `json:"processType,omitempty"`
```
drgarcia1986: Why is it a pointer?

k8s/deployments.go
```diff
@@ -378,7 +378,11 @@ func newWelcomeContainer(app *models.App) (c *api.Container) {
 func newSlugRunnerContainer(app *models.App, slug string, storageType string) (c *api.Container) {
 	// creating runner container
 	c = newContainer(*app.Name, slugRunnerImage)
-	c.Args = []string{"start", "web"}
+	processType := processTypeWeb
+	if app.ProcessType != nil {
+		processType = *app.ProcessType
+	}
+	c.Args = []string{"start", processType}
```
drgarcia1986: Is it really necessary to check this value here?

k8s/apps.go
alexguerra-luizalabs: Sure, my bad.

k8s/deployments.go
alexguerra-luizalabs: It is not, if we edit old apps and put the process type in the annotation this check is useless.
The question is: edit old apps or not? Not sure....I supposed we won't edit but editing is not a big deal either.

models/app_in.go
alexguerra-luizalabs: go-swagger's auto generated code.

k8s/deployments.go
drgarcia1986: Good point!
Wen can remove this check after migration and we don't need to dealing with old apps.

Any plans to "fix" this lost information in old apps?

models/app_in.go
drgarcia1986: 😐 🔫

k8s/deployments.go
alexguerra-luizalabs: I'm changing my mind to remove this check and force the editing otherwise this check will stay forever.

drgarcia1986: Sounds good.
We can work together to fix it.

PR #169, opened 2017-03-22
Read keys from k8s secrets

aguerra:
:horse:

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/169)


handlers/auth.go
```diff
@@ -18,24 +18,38 @@ import (
...
 )
 
+func getKeyBytes(keyName string) ([]byte, error) {
+	s, err := k8s.Client.Secrets().Get(namespace, secretName)
+	if err != nil {
+		return nil, err
+	}
+	data, ok := s.Data[keyName]
+	if !ok {
+		return nil, fmt.Errorf("Missing key %s", keyName)
```
drgarcia1986: Maybe create a error variable is more idiomatic, like `var ErrMissingKey = errors.New(fmt.Sprintf("Missing key %s", keyName))`, What do you think?

aguerra: Didn't know that, I thought the shorter form was preferred. So:
```
func ....
    var errMissingKey = errors.New(fmt.Sprintf("Missing key %s", keyName))
    ...
    if !ok {
      return nil, errMisingKey
    } 
```
?

handlers/auth.go
```diff
@@ -18,24 +18,40 @@ import (
...
+	secretName  = "teresa-keys"
+	namespace   = "kube-paas"
 )
 
 var (
 	verifyKey *rsa.PublicKey
 	signKey   *rsa.PrivateKey
 )
 
+var ErrMissingKey = std_errors.New("Missing key")
```
drgarcia1986: What do you think about create it var inside previous var list ?

handlers/auth.go
```diff
@@ -2,7 +2,7 @@ package handlers
 
 import (
 	"crypto/rsa"
-	"io/ioutil"
+	std_errors "errors"
```
drgarcia1986: Why?

aguerra: There's a `import "github.com/go-openapi/errors"` below.

handlers/auth.go
aguerra: :+1:

handlers/auth.go
drgarcia1986: 😢

Issue #151, opened 2017-01-12
Client times out but deploy continues in background


alexguerra-luizalabs:
Mainly with java applications.

aguerra:
Fixed by PR #174.

Issue #139, opened 2016-12-27
Remove hardcoded timeouts


alexguerra-luizalabs:
Mainly from the build process.

aguerra:
Fixed by PR #174.

PR #170, opened 2017-03-30
Increase deploy timeouts and make them configurable

aguerra:
Also upgrade to the latest envconfig.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/170)


k8s/deployments.go
```diff
@@ -59,13 +59,15 @@ type deploy struct {
...
+	FinishTimeout        time.Duration `split_words:"true" default:"30m"`
+	RevisionHistoryLimit int32         `split_words:"true" default:"5"`
+	StartTimeout         time.Duration `split_words:"true" default:"10m"`
 }
 
 var deploymentConfig DeploymentConfig
 
 func init() {
-	err := envconfig.Process("teresadeploy", &deploymentConfig)
+	err := envconfig.Process("teresa_deploy", &deploymentConfig)
```
drgarcia1986: Why?
Is it because the upgrade of `envconfig` dependency ?

drgarcia1986:
Related #139 #151

k8s/deployments.go
aguerra: No, I prefer env vars `TERESA_* style.

drgarcia1986: Hum 🤔 but, this change will broke the env `TERESADEPLOY_REVISION_HISTORY_LIMIT` and is out of pattern of others env configs.

What do you think about?

aguerra: I would changed it right now as it is undocumented, but to be on the safer side we can change the env vars style for `0.3.0` and document the change. Agree?

drgarcia1986: It's "documented" on README...

But :shipit:

PR #172, opened 2017-03-31
define quota limits with some default values

drgarcia1986:
Being objective. For every app created on k8s cluster via teresa, we needing to change the app quote by hand and we found it embarrassing.

We are going to start a big change in both (CLI and API) and this simple change will help us to work hard for it.

This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/172)


drgarcia1986:
We'll discuss about it.

PR #174, opened 2017-04-18
Fix deploy timeouts

aguerra:
Deploy is a streaming operation and we need the connection open even with the presence of firewalls or load balancers that have idle timeouts. As we don't control go-swagger autogenerated server we send a "ping" to the client every 30 seconds by default (this value may be changed using the TERESADEPLOY_KEEPALIVE_TIMEOUT env var).

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/174)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [NEXT_RELEASE]
+### Fixed
+- Fix deploy timeouts by sending a few bytes at a constant time interval to the
+  network connection in use (as required by the idle timeout of the aws classic
+  elb for instance)
+
 ## [0.3.0] - 2017-04-07
 ### Added
 - Support for non-web process types
--- a/k8s/deployments.go
+++ b/k8s/deployments.go
@@ -26,6 +26,10 @@ import (
 	"k8s.io/kubernetes/pkg/util/wait"
 )
 
+const (
+	keepAliveMessage = "\u200B" // Zero width space
+)
+
 // DeploymentsInterface is used to allow mock testing
 type DeploymentsInterface interface {
 	Deployments() DeploymentInterface
@@ -59,6 +63,7 @@ type deploy struct {
 }
 
 type DeploymentConfig struct {
+	KeepaliveTimeout     time.Duration `split_words:"true" default:"30s"`
 	FinishTimeout        time.Duration `split_words:"true" default:"30m"`
 	RevisionHistoryLimit int32         `split_words:"true" default:"5"`
 	StartTimeout         time.Duration `split_words:"true" default:"10m"`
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [NEXT_RELEASE] ##
```
drgarcia1986:
```diff
- ## [NEXT_RELEASE] ##
+ ## [NEXT_RELEASE]
```

k8s/deployments.go
```diff
@@ -98,60 +158,27 @@ func (c deployments) Create(appName, description string, file *runtime.File, sto
...
+		for {
+			select {
+			case <-time.After(deploymentConfig.KeepAliveTimeout):
+				fmt.Fprint(w, keepAliveMessage)
+			case <-done:
+				break L
+			}
 		}
-		fmt.Fprintln(w, "deploy finished with success")
+		w.Close()
```
drgarcia1986: Why not `defer w.Close()` on start of anonymous function?

k8s/deployments.go
```diff
@@ -98,60 +158,27 @@ func (c deployments) Create(appName, description string, file *runtime.File, sto
...
-		lc.Debug("build step done without errors")
-		fmt.Fprintln(w, "build step done without errors")
-		// updating deployment
-		lc.Debug("updating deployment for rolling update...")
-		fmt.Fprintln(w, "rolling update...")
-		if err := c.updateDeployment(app, deploy.slugPath, deploy.description, storage); err != nil {
-			lc.WithError(err).Error("error updating deployment")
-			fmt.Fprintln(w, "error when doing rolling update")
-			return
+	L:
```
drgarcia1986: What do you think about use `Loop:` instead of `L:` for this label ?
Maybe it will help us for a quick understand of this block of code

k8s/deployments.go
```diff
@@ -98,60 +158,27 @@ func (c deployments) Create(appName, description string, file *runtime.File, sto
...
+		for {
+			select {
+			case <-time.After(deploymentConfig.KeepAliveTimeout):
+				fmt.Fprint(w, keepAliveMessage)
+			case <-done:
+				break L
```
drgarcia1986: or you can simplify it using two `break`, one for `select` and another for `for`.

CHANGELOG.md
aguerra: :+1:

k8s/deployments.go
aguerra: I'm going to pick two breaks.

k8s/deployments.go
aguerra: Good point :+1:

k8s/deployments.go
```diff
@@ -90,6 +95,61 @@ func (c deployments) CreateWelcomeDeployment(app *models.App) error {
...
+	storage helpers.Storage,
+	done chan<- struct{},
+) {
+	// create log context with uuid and app name
+	lc := log.WithField("app", app.Name).WithField("deployUUID", deploy.uuid)
+	lc.Info("starting deploy...")
+	defer close(done)
+
+	// parsing teresa.yaml and updating app
+	if err := parsingTeresaYaml(app, file.Data); err != nil {
```
arnaldopereira: If parsing `teresa.yaml` fails, the deployment goroutine shouldn't be created in the first place. Maybe we should try and parse it before creating the deployment goroutine and pass the `file.Data` to the deployment goroutine.

Also, the name of the function that would parse the `teresa.yaml` would be something as `parseTeresaYaml` instead of `parsingTeresaYaml`.

arnaldopereira: As a side effect, we would decrease the this `create`s code and logic, which would be nice.

arnaldopereira: If you guys agree with that, please don't forget to add a `defer` closing the `file.Data` outside this goroutine - and sorry for the many comments in series.

aguerra: Maybe you have just fixed a strange periodic S3 related bug with this last suggestion :+1:

PR #175, opened 2017-04-18
change KeepAlive time from 3 minutes to 30 seconds

drgarcia1986:
Another try to fix client timeouts.

This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/175)


drgarcia1986:
doesn't work 😭

PR #182, opened 2017-04-26
big gRPC change WIP

drgarcia1986:
still in `WIP`

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/182)


pkg/server/server.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil, err
+	}
+
+	s := grpc.NewServer(grpc.Creds(creds))
+
+	cs := user.NewServer(user.NewDatabaseOperations(opt.DB, opt.RSAPrivateKey))
+	cs.RegisterServer(s)
+
+	return &Server{listener: l, grpcServer: s}, nil
+}
```
aguerra: Maybe use envconfig here (or in main) to configure TLSCert, TLSKey

pkg/client/client.go
```diff
@@ -0,0 +1,30 @@
...
+
+func New(cfg ClusterConfig) (*grpc.ClientConn, error) {
+	tlsConfig := &tls.Config{InsecureSkipVerify: true}
+	creds := credentials.NewTLS(tlsConfig)
+	return grpc.Dial(
+		cfg.Server,
+		grpc.WithPerRPCCredentials(&tokenAuth{cfg.Token}),
+		grpc.WithTransportCredentials(creds),
+	)
+}
```
aguerra: Are `GetRequestMetadata` `RequireTransportSecurity` grpc stuff?

cmd/client/cmd/login_cluster.go
```diff
@@ -2,12 +2,20 @@ package cmd
...
+	context "golang.org/x/net/context"
 
-	"github.com/go-openapi/strfmt"
 	"github.com/howeyc/gopass"
+	"github.com/luizalabs/teresa-api/cmd/client/connection"
+	"github.com/luizalabs/teresa-api/pkg/client"
 	"github.com/spf13/cobra"
+
+	userpb "github.com/luizalabs/teresa-api/pkg/protobuf"
 )
```
aguerra: Forget it, userpb is ok.

drgarcia1986:
DON'T MERGE!

pkg/client/client.go
drgarcia1986: Yep, from the [PerRPCCredentials](https://godoc.org/google.golang.org/grpc/credentials#PerRPCCredentials) interface.

pkg/server/server.go
drgarcia1986: `main.go` it's the right answer (I think), but, I don't know how we'll deal with TLSCerts (on secrets, I know, but, in files or bytes?)

pkg/server/user/handlers.go
```diff
@@ -0,0 +1,29 @@
...
+	return &userpb.LoginResponse{Token: token}, nil
+}
+
+func (s *Server) RegisterServer(grpcServer *grpc.Server) {
+	userpb.RegisterUserServer(grpcServer, s)
+}
+
+func NewServer(ops Operations) *Server {
+	return &Server{ops: ops}
+}
```
drgarcia1986: Agree

drgarcia1986:
Agree with "pkg common" comment, but, are you have any suggestion for that case?

cmd/client/cmd/login_cluster.go
drgarcia1986: You're right `protobuf` is the common package for all protobuf, but, I just don't like `pb` as package name on big code bases 😂

aguerra: :+1:

pkg/server/server.go
aguerra: I prefer something like `export TLS_CERT=/etc/server.crt`

pkg/client/client.go
aguerra: :+1:

pkg/server/user/handlers.go
drgarcia1986: `:GoRename` ❤️

aguerra:
For now I would move `pkg/common/errors.go -> pkg/errors/errors.go`, but you are right, it will not be always possible to avoid common.

drgarcia1986:
If we do this, we need to deal with conflicts with standard library (`errors` package)

pkg/server/server.go
drgarcia1986: LGTM, but are you thinking to change RSA key schema? 
I mean, we need to deal in the same way for both secrets types.

aguerra: I would mount secrets as files and only deal with files on code, but would postpone this change if I couldn't do it in all places (too big, too risky, etc).

drgarcia1986:
@aguerra to avoid `common` pkg, what about move `ErrPermisionDenied` to `server/user` ? We're talking about users right?

PR #185, opened 2017-05-04
Add gRPC support for user Login and SetPassword method

drgarcia1986:
This is the first PR of gRPC series 🎉 

**Includes**:
- New project file tree (`/pkg/`, `/cmd/`, etc. Please, pay attention)
- New interfaces 🚀 
- protobufs (only for `User` domain for now)
- gRPC stuff (server, user service, client, etc)
- Unit tests 🚀  (`go test $(go list ./... | grep -v vendor)`) (obs: tests of pkg `github.com/luizalabs/teresa-api/pkg/server/secrets` failed because some external dependency is started in `init()` functions)
- New command `set-password` to change the password for the current user.
- and much more.

**Need help**:
- [x] Lovely and affectionate review ❤️ 
- [x] Update vendor/dependencies
- [x] ~Run server with go-swagger mux and gRPC server on the same port.~
- [x] Define the official format to handle secrets (e.g. RSA Keys, TLS Certs, etc.) (either K8S API or OS files)
- [x] Run integration tests for each changed command (`login` and `set-password`)
- [ ] Some logical duplicated (but maybe we need to deal with it for now)

**Before merge**:
- [x] Remove old `login` stuff (go-swagger handler, swagger definition)
- [ ] Remove "config-file" duplicated stuff 
- [x] Remove gRPC cli logs
- [x] Remove/Refact URL parser/validator of method `set-cluster` of cli.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/185)


pkg/client/client.go
```diff
@@ -0,0 +1,30 @@
...
+
+func New(cfg ClusterConfig) (*grpc.ClientConn, error) {
+	tlsConfig := &tls.Config{InsecureSkipVerify: true}
+	creds := credentials.NewTLS(tlsConfig)
+	return grpc.Dial(
+		cfg.Server,
+		grpc.WithPerRPCCredentials(&tokenAuth{cfg.Token}),
+		grpc.WithTransportCredentials(creds),
+	)
+}
```
drgarcia1986: You're right... From [doc/comments](https://godoc.org/google.golang.org/grpc/credentials#PerRPCCredentials):
```
    // RequireTransportSecurity indicates whether the credentials requires
    // transport security
```
I made a dummy test with `RequireTransportSecury = true` on client without TLS transport and raise this error:
```
the credentials require transport level security (use grpc.WithTransportCredentials() to set)
```
I think we can change to `true` 👍

PR #186, opened 2017-05-05
Fix multi teams tokens by avoiding pointer to range var

aguerra:
Before this fix a user would get access only to apps from one of its teams.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/186)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [NEXT_RELEASE]
 ### Fixed
 - Finish the merge with `teresa-cli` by removing all references to the old repo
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [NEXT_RELEASE]
 ### Fixed
 - Finish the merge with `teresa-cli` by removing all references to the old repo
+- Fix multi team tokens. Before this fix a user would get access only to apps
```
drgarcia1986: you can remove `fix` word (already on `fixed` section)

aguerra: :+1:

Issue #203, opened 2017-06-02
Use gRPC instead of go-swagger 


drgarcia1986:
### Macro vision checklist:
- [x] Login ( #185 )
- [x] create user ( #192 #200 )
- [x] delete user ( #190 )
- [x] create team ( #189 )
- [x] team list ( #197 )
- [x] team add-user ( #193 )
- [x] app create ( #207  )
- [x] app env-set ( #213 )
- [x] app env-unset ( #213 )
- [x] app info (#208)
- [x] app list ( #244 )
- [X] app logs ( #205 )
- [x] deploy ( #211  )
----------
### Extras:
- [x] set-password command ( #185 )
- [x] [server] create super user command ( #198 )
- [x] TLS (some PRs of series) 
- [x] Release command phase ( #226 )
- [x] Health Check ( #230 )
- [x] Version by tag (git) ( #229 )
- [x] minio slug storage ( #236 )
---------
### Refactorings:
- [x] cli configs ( #188 )
- [x] bash completion ( #187 ) 
- [x] delete build pods after work ( #211 )
- [x] standardize errors msgs format and server logs (server and client) ( #214 #218  )
---------
### Fun Stuff
- [X] Unit Tests
- [X] Travis CI (run unit tests against some golang versions ) ( #194 )

drgarcia1986:
The command `team list` doesn't returns the team members list if called for a non-admin user.

drgarcia1986:
🎉

PR #211, opened 2017-07-12
App Deploy (gRPC series)

drgarcia1986:
related #203 

It's a big pull request with tons of code change on sensible features, please, pay attention and don't hesitate to point errors or even improvements.

BTW, now Teresa will delete "build pods" after they finish their job.

Good Luck 🇯🇲

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/211)


pkg/server/deploy/deploy.go
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+	hasSrv, err := ops.k8s.HasService(a.Name, a.Name)
+	if err != nil {
+		log.WithError(err).Errorf("Checking APP service")
+		return
+	}
+	if !hasSrv {
+		fmt.Fprintln(w, "Exposing LoadBalancer service")
+		if err := ops.k8s.CreateService(a.Name, a.Name); err != nil {
```
aguerra: There are also `ClusterIP` and `NodePort` services.

pkg/server/deploy/handlers.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type Service struct {
+	ops     Operations
+	options *Options
+}
+
+func (s *Service) Make(stream dpb.Deploy_MakeServer) error {
+	var appName, description string
+	content := new(bytes.Buffer)
```
aguerra: Maybe we should a use a temporary file for the content because these tar files are big and keeping them in memory may increase the memory requirements.

pkg/server/deploy/handlers.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func sendMsgsToAChannel(r io.Reader) <-chan string {
+	c := make(chan string)
+	go func() {
+		defer close(c)
+		scanner := bufio.NewScanner(r)
+		for scanner.Scan() {
+			c <- fmt.Sprintln(scanner.Text())
+		}
```
aguerra: I guess we should check scanner.Err() at the end of the loop, not sure though.

pkg/server/deploy/handlers.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	rs := bytes.NewReader(content.Bytes())
+	rc, err := s.ops.Deploy(u, appName, rs, description, s.options.RevisionHistoryLimit)
+	if err != nil {
+		return err
+	}
+	defer rc.Close()
+
+	deployMsgs := sendMsgsToAChannel(rc)
```
aguerra: Maybe rename `sendMsgsToAChannel -> channelFromReader`

pkg/server/k8s/client.go
```diff
@@ -312,16 +311,133 @@ func (k *k8sClient) Limits(namespace, name string) (*app.Limits, error) {
...
+			return
+		}
+
+		stream, err := k.PodLogs(podSpec.Namespace, podSpec.Name, 10, true)
+		if err != nil {
+			return
+		}
+		io.Copy(w, stream)
+
+		if err = k.waitPodEnd(pod, 1*time.Second, 5*time.Minute); err != nil {
```
aguerra: For deploy safety we need more than 5 minutes of finish timeout, maybe more args. Also we may increase the poll interval (less cpu intensive)

pkg/server/k8s/client.go
```diff
@@ -312,16 +311,133 @@ func (k *k8sClient) Limits(namespace, name string) (*app.Limits, error) {
...
+	if err != nil {
+		return 1
+	}
+	for _, containerStatus := range p.Status.ContainerStatuses {
+		state := containerStatus.State.Terminated
+		if state.ExitCode != 0 {
+			return int(state.ExitCode)
+		}
+	}
+	return 0
```
aguerra: Maybe return error if the pod is running

pkg/server/k8s/client.go
```diff
@@ -312,16 +311,133 @@ func (k *k8sClient) Limits(namespace, name string) (*app.Limits, error) {
 	return lim, nil
 }
 
-func newInClusterK8sClient() (Client, error) {
-	conf, err := restclient.InClusterConfig()
+func (k *k8sClient) CreateDeploy(deploySpec *deploy.DeploySpec) error {
+	replicas := k.currentPodReplicasFromDeploy(deploySpec.Namespace, deploySpec.Name)
```
aguerra: Maybe a django style name `CreateOrUpdateDeploy`

cmd/client/cmd/deploy.go
```diff
@@ -209,3 +151,116 @@ func init() {
...
+	for {
+		buf := make([]byte, 1024)
+		n, err := r.Read(buf)
+		if err != nil && err != io.EOF {
+			fmt.Fprintln(os.Stderr, "Error reading bytes of temp file:")
+			return err
+		}
+		if n == 0 {
+			break
+		}
```
drgarcia1986: Yeah, much better.

pkg/server/app/app.go
```diff
@@ -19,6 +19,9 @@ type Operations interface {
 	Create(user *storage.User, app *App) error
 	Logs(user *storage.User, appName string, lines int64, follow bool) (io.ReadCloser, error)
 	Info(user *storage.User, appName string) (*Info, error)
+	TeamName(appName string) (string, error)
+	Meta(appName string) (*App, error)
+	HasPermission(user *storage.User, appName string) bool
```
drgarcia1986: Me too.
I'll think again about this name, but for now, I don't have any idea.

pkg/server/deploy/deploy.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	deployId := genDeployId()
+	buildDest := fmt.Sprintf("deploys/%s/%s/out", appName, deployId)
+
+	r, w := io.Pipe()
+	go func() {
+		defer w.Close()
+		if err = ops.buildApp(tarBall, a, deployId, buildDest, w); err != nil {
+			return
```
drgarcia1986: Oh, I forgot :/

pkg/server/app/app.go
drgarcia1986: Let me know if you find a better name.

pkg/server/deploy/handlers.go
drgarcia1986: 😱 🏆 This name is much better on a monumental scale.

pkg/server/deploy/handlers.go
drgarcia1986: If raise an error we can't do much more than log the error :/

aguerra: fair enough

pkg/server/k8s/client.go
drgarcia1986: I don't like functions with more than one proposal but you're right, the current name does not reflect the reality.

pkg/server/k8s/client.go
drgarcia1986: Hm, What do you think about 30 minutes for finish timeout and 3 seconds for check interval?

pkg/server/k8s/client.go
drgarcia1986: Do you say that because `Terminated` is a pointer and it can be nil?
I'll prevent this.

pkg/server/k8s/client.go
aguerra: That's ok.

pkg/server/deploy/handlers.go
drgarcia1986: I disagree with a temporary file because we need to "untar" content, look for teresa.yaml file and send to s3 (as an `io.ReadSeeker`), in other words, if we put the content in a temporary file and erase buffer, we'll need to read it again (in memory I suppose).

I know, these tar files are big (maybe huge). I didn't find a good way to deal with their though.

pkg/server/k8s/client.go
aguerra: If I call this function on a running pod I would expect an error: pod is still running, no exit code yet. But this is :nail_care:

drgarcia1986: Gotcha!!! I'll change

pkg/server/deploy/handlers.go
aguerra: Ok, I guess as a last resort it's worth to take a look at the old code, it uses http://godoc.org/github.com/go-openapi/runtime#File and https://golang.org/pkg/mime/multipart/#File to apparently not put everything in memory, but I'm just guessing, if it can't be done we should move on.

pkg/server/deploy/deploy.go
drgarcia1986: Yes, you're right but today we only deal with `LoadBalancer` service, right?
I not sure if we need to deal with different types of service at this moment, what do you think?

The function `pkg/server/k8s/converters.go/serviceSpec` has an argument called `srvType` and uses it on service specification, We'll need to delegate that information (service type of app) to the app layer though

aguerra: Sure, lets keep ""Exposing LoadBalancer service"

pkg/server/deploy/handlers.go
drgarcia1986: I took a look at the old code but is a lot of magic :/ (i.e. go-swagger magic).
I'll try to understand how to work with `runtime.File` and `multipart.File` on gRPC context.

pkg/server/app/app.go
drgarcia1986: What do you think about `Get` ? 
I'm not sure because that method returns only information about the app, not a powerful struct with useful methods. BUT, is better than Meta? (or not? 🤔 )

aguerra: It' better, perhaps `Instance`.

pkg/server/deploy/handlers.go
drgarcia1986: For now, we won't change that logic.

PR #220, opened 2017-07-20
create a custom writer to use with logs of grpc

drgarcia1986:
Why? The aws elb performs health-checks with HTTP/1 protocol and without TLS certificates. It makes a mess on Teresa logs. To avoid silencing logs completely I created a simple mechanism to filter some log messages. Let me know what you think.

This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/220)

### DIFF
--- a/pkg/server/log.go
+++ b/pkg/server/log.go
@@ -1,23 +1,23 @@
 package server
 
 import (
+	"bytes"
 	"io"
 	"log"
-	"strings"
 )
 
 type FilteredWriter struct {
 	io.Writer
 }
 
-var msgsToFilter = []string{
-	"http2Server.HandleStreams failed to receive the preface from client",
-	"failed to complete security handshake from",
+var msgsToFilter = [][]byte{
+	[]byte("http2Server.HandleStreams failed to receive the preface from client"),
+	[]byte("failed to complete security handshake from"),
 }
 
 func (f *FilteredWriter) Write(b []byte) (int, error) {
 	for _, msg := range msgsToFilter {
-		if strings.Contains(string(b), msg) {
+		if bytes.Contains(b, msg) {
 			return 0, nil
 		}
 	}
--- a/pkg/server/log.go
+++ b/pkg/server/log.go
@@ -1,0 +1,29 @@
+package server
+
+import (
+	"io"
+	"log"
+	"strings"
+)
+
+type FilteredWriter struct {
+	io.Writer
+}
+
+var msgsToFilter = []string{
+	"http2Server.HandleStreams failed to receive the preface from client",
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -67,6 +69,7 @@ func New(opt Options) (*Server, error) {
 	}
 
 	s := grpc.NewServer(sOpts...)
+	grpclog.SetLogger(NewLogger(os.Stdout))
 
 	us := user.NewService(uOps)
 	us.RegisterService(s)
```
drgarcia1986: Makes senses, but, I disagree with toil.
Change elb configs by hand is so annoying and isn't multi-cloud.

drgarcia1986: Is there a form to do that by k8s annotation?

aguerra: I don't think so.

pkg/server/log.go
```diff
@@ -0,0 +1,29 @@
...
+	"io"
+	"log"
+	"strings"
+)
+
+type FilteredWriter struct {
+	io.Writer
+}
+
+var msgsToFilter = []string{
```
arnaldopereira: Wouldn't this hide real errors parsing ssl certificates from client?

drgarcia1986: Maybe, but only on server side.
IMO, that log message (`remote error: tls: bad certificate`) doesn't help too much.

arnaldopereira: If teresa-server is flooded with a whole lot of HTTP requests without TLS, this change would hide that from us.

We should try to fix the ELB behavior instead.

drgarcia1986: If the solution for that annoying problem is specific for AWS ELB, I disagree. As I said [here](https://github.com/luizalabs/teresa-api/pull/220#discussion_r128630487), however, I'll search a generic fix.

drgarcia1986: I made a mistake, the correct log message of elb health check is `grpc: Server.Serve failed to complete security handshake from "<ip>": EOF` and not `tls: bad certificate`... sorry, I'll fix.

drgarcia1986:
@aguerra @arnaldopereira guys I didn't find any way to deal with the aws ELB in a generic form, so, What do you think about it? Either ignore this PR and `grep -v` FTW, or, `rebase and merge`?

drgarcia1986:
maybe with the #230 we don't need this.

arnaldopereira:
Now we're talking. 😀

drgarcia1986:
we don't need it.
the problem is fixed, take a look at #234 and #230.

PR #221, opened 2017-07-20
Implement versioning scheme based on ldflags

aguerra:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/221)


Makefile
```diff
@@ -64,8 +66,11 @@ shell:
 run-api-server:
 	go run ./cmd/server/main.go run --port 8080
 
+build-server:
+	@go build -ldflags "-X $(TERESA_API)/pkg/version.Version=$(TERESA_VERSION)" -o teresa-server $(TERESA_API)/cmd/server
+
 build-client:
-	@go build -o teresa cmd/client/main.go
+	@go build -ldflags "-X $(TERESA_API)/pkg/version.Version=$(TERESA_VERSION)" -o teresa $(TERESA_API)/cmd/client
```
drgarcia1986: :clap:

Makefile
```diff
@@ -64,8 +66,11 @@ shell:
 run-api-server:
 	go run ./cmd/server/main.go run --port 8080
 
+build-server:
```
drgarcia1986: There are a few differences in `build-server` command between Makefile and Dockerfile, what do you think about unify?

Makefile
```diff
@@ -3,6 +3,8 @@ IMAGE_NAME=teresa
 IMAGE_VERSION=0.1.1
 IMAGE_INSTANCE=default
 TERESA_DOCKER_PORT ?= 8080
+TERESA_VERSION ?= latest
```
drgarcia1986: We can use current commit hash, will be awesome 😄

Makefile
aguerra: Sure, I was thinking of doing last.

PR #227, opened 2017-07-31
No TLS by default

drgarcia1986:
To simplify the up and run process for new users, this PR changes the default behavior of server and client to **don't** use TLS by default.

Today we need to use the flag `--notls` for each command. With this PR you only need to use the flag `--tls` when you use the `set-cluster`  command if you want to use TLS.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/227)

### DIFF
--- a/cmd/client/cmd/config.go
+++ b/cmd/client/cmd/config.go
@@ -110,9 +110,18 @@ func setCluster(cmd *cobra.Command, args []string) {
 	if server == "" {
 		client.PrintErrorAndExit("Server URI not provided")
 	}
-	useTLS, _ := cmd.Flags().GetBool("tls")
-	insecure, _ := cmd.Flags().GetBool("tlsinsecure")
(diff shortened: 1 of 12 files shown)


cmd/client/cmd/config.go
```diff
@@ -109,9 +106,13 @@ func setCluster(cmd *cobra.Command, args []string) {
...
 		return
 	}
-	if serverFlag == "" {
+	server, _ := cmd.Flags().GetString("server")
+	if server == "" {
 		client.PrintErrorAndExit("Server URI not provided")
 	}
+	useTLS, _ := cmd.Flags().GetBool("tls")
+	insecure, _ := cmd.Flags().GetBool("tlsinsecure")
+	current, _ := cmd.Flags().GetBool("current")
```
aguerra: if err exit

drgarcia1986: done 👍

PR #230, opened 2017-08-01
expose a simple health check on a http server

drgarcia1986:
Useful feature to monitoring teresa and k8s health outside of cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/230)


drgarcia1986:
With this PR we don't need the filter of #220 .

PR #231, opened 2017-08-01
[Proposal] Make slug images configurable by env vars

aguerra:
We should be able to change the images fast (for security patches for instance) and without a new deploy.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/231)


aguerra:
updated @drgarcia1986

PR #232, opened 2017-08-01
[server] upgrade golang version to 1.8

drgarcia1986:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/232)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.6
+FROM golang:1.8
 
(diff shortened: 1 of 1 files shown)


drgarcia1986:
@aguerra I'm not sure, because the gRPC from vendor uses `x/context`, I'll check

drgarcia1986:
To tell the truth, that problem isn't on the vendor but is on the generated protos.

PR #235, opened 2017-08-02
Fix teresa client root documentation

aguerra:
Fixes, better style and document new flags.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/235)

### DIFF
--- a/cmd/client/cmd/root.go
+++ b/cmd/client/cmd/root.go
@@ -19,12 +19,11 @@ var log *logrus.Logger
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
 	Use:   "teresa",
-	Short: "Teresa cli",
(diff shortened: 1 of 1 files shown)


cmd/client/cmd/root.go
```diff
@@ -33,7 +32,13 @@ one to use.
 
 To set a cluster, eg.:
 
-  $ teresa config set-cluster my_cluster_name -s https://mycluster.mydomain.com
+  $ teresa config set-cluster my_cluster_name -s mycluster.mydomain.com
```
drgarcia1986: I'm not sure but I think the flag `-s` doesn't works anymore.

aguerra: You are right!

aguerra: Fixed.

PR #237, opened 2017-08-03
Return list of users in `team list` command for a non admin users

drgarcia1986:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/237)

### DIFF
--- a/pkg/server/team/team.go
+++ b/pkg/server/team/team.go
@@ -71,13 +71,8 @@ func (dbt *DatabaseOperations) List() ([]*storage.Team, error) {
 		)
 	}
 
-	for _, t := range teams {
-		if err := dbt.DB.Model(t).Association("Users").Find(&t.Users).Error; err != nil {
-			return nil, teresa_errors.New(
-				teresa_errors.ErrInternalServerError,
-				errors.Wrap(err, fmt.Sprintf("associating team %s with its users", t.Name)),
-			)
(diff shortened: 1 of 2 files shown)


pkg/server/team/team.go
```diff
@@ -95,9 +90,25 @@ func (dbt *DatabaseOperations) ListByUser(userEmail string) ([]*storage.Team, er
...
 		)
 	}
+
+	if err = dbt.associateUsersWithTeams(teams); err != nil {
+		return nil, err
+	}
 	return teams, nil
 }
 
+func (dbt *DatabaseOperations) associateUsersWithTeams(teams []*storage.Team) error {
```
arnaldopereira: I don't like this function name. It associates users with teams and perform a find operation, so it seems it goes a _little_ beyond the association.

drgarcia1986: hm you're right @arnaldopereira .
What about `findUsersOfTeams` ?

aguerra: Maybe `findTeamUsers`?

drgarcia1986: ☝️ LGTM

arnaldopereira: LGTM

drgarcia1986:
updated @aguerra @arnaldopereira

PR #239, opened 2017-08-03
Fix app log lost connections

aguerra:
By always sending a report if there are logs or not.
Also useful with --follow.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/239)

### DIFF
--- a/pkg/goutil/channels.go
+++ b/pkg/goutil/channels.go
@@ -1,0 +1,25 @@
+package goutil
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/goutil/channels.go
```diff
@@ -0,0 +1,25 @@
...
+	go func() {
+		defer close(c)
+		scanner := bufio.NewScanner(r)
+		for scanner.Scan() {
+			c <- fn(scanner.Text())
+		}
+	}()
+
+	return c
+}
```
drgarcia1986: What do you think about write some simple tests for this method ?

aguerra: Done!

PR #241, opened 2017-08-04
Make create app delete the namespace on errors

aguerra:
Also emit better error messages.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/241)

### DIFF
--- a/pkg/server/app/app.go
+++ b/pkg/server/app/app.go
@@ -46,6 +46,7 @@ type K8sOperations interface {
 	SetNamespaceAnnotations(namespace string, annotations map[string]string) error
 	DeleteDeployEnvVars(namespace, name string, evNames []string) error
 	CreateOrUpdateDeployEnvVars(namespace, name string, evs []*EnvVar) error
+	DeleteNamespace(namespace string) error
 }
 
 type AppOperations struct {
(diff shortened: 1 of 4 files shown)


pkg/server/app/app.go
```diff
@@ -96,8 +97,14 @@ func (ops *AppOperations) Create(user *storage.User, app *App) error {
 		return teresa_errors.NewInternalServerError(err)
 	}
 
+	defer func() {
+		if Err != nil {
```
drgarcia1986: ⚠️  Be careful! you need to check if error is different of `ErrAlreadyExists` or `ErrPermissionDenied` before perform the `delete namespace` action

aguerra: Very good point but I guess the defer is only installed if me passed the stage of returning these 2 errors, what do you think?

drgarcia1986: You're right, but you know, as we say in Portuguese "o seguro morreu de velho". Could you add a test to prevent this fearful behavior ?

arnaldopereira: +1 for being extra careful with this

aguerra: @drgarcia1986 Added tests.

PR #248, opened 2017-08-09
Add min password length to the set-password command

aguerra:
Better security and to avoid mistakes.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/248)

### DIFF
--- a/pkg/client/password.go
+++ b/pkg/client/password.go
@@ -6,6 +6,17 @@ import (
 	"github.com/howeyc/gopass"
 )
 
+const (
+	minPassLen = 8
+)
+
+func checkPassword(pass string) error {
+	if len([]rune(pass)) < minPassLen {
+		return fmt.Errorf("minimum password length is %d", minPassLen)
+	}
+	return nil
+}
+
 func GetMaskedPassword(prompt string) (string, error) {
 	fmt.Print(prompt)
 	p, err := gopass.GetPasswdMasked()
@@ -13,5 +24,8 @@ func GetMaskedPassword(prompt string) (string, error) {
 		return "", err
 	}
 	fmt.Print("\r")
+	if err := checkPassword(string(p)); err != nil {
+		return "", err
+	}
 	return string(p), nil
 }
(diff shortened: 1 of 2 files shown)


pkg/client/password.go
```diff
@@ -6,12 +6,26 @@ import (
...
+const (
+	minPassLen = 8
+)
+
+func checkPassword(pass string) error {
+	if len([]rune(pass)) < minPassLen {
+		return fmt.Errorf("minimum length is %d", minPassLen)
+	}
+	return nil
+}
```
drgarcia1986: What do you think about writing a test for this function?

pkg/client/password.go
```diff
@@ -6,12 +6,26 @@ import (
 	"github.com/howeyc/gopass"
 )
 
+const (
+	minPassLen = 8
+)
+
+func checkPassword(pass string) error {
+	if len([]rune(pass)) < minPassLen {
+		return fmt.Errorf("minimum length is %d", minPassLen)
```
drgarcia1986: maybe `minimum password length is...`

pkg/client/password.go
aguerra: @drgarcia1986 updated. I guess creating specific errors in this case is not worthwhile so I only checked for error and success.

pkg/client/password_test.go
```diff
@@ -0,0 +1,21 @@
...
+		pass string
+		ok   bool
+	}{
+		{"12345678", true},
+		{"test", false},
+	}
+
+	for _, tc := range testCases {
+		if err := checkPassword(tc.pass); err != nil && tc.ok {
+			t.Errorf("expected failure, got success")
```
drgarcia1986: 💅  I'm bit confused here, please don't get me wrong, right?
What do you think about changing for something like this:
```
if err := checkPassword(tc.in); (err != nil) && tc.expectError {
    t.Errorf("expected error: %v, got error %v", tc.expectError, err)
} 
```
💅

aguerra: @drgarcia1986 updated, I think the original logic is wrong, we have to check both cases.

PR #249, opened 2017-08-09
Better Dockerfile

aguerra:
- use default grpc port
- use ENTRYPOINT, which is more flexible: we can vary the command
  parameters without rebuilding the image

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/249)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,5 +6,6 @@ COPY . /go/src/github.com/luizalabs/teresa-api
 
 RUN make build-server
 
-CMD ["./teresa-server", "run", "--port", "8080"]
-EXPOSE 8080
+ENTRYPOINT ["./teresa-server"]
+CMD ["run"]
+EXPOSE 50051


Dockerfile
```diff
@@ -6,5 +6,5 @@ COPY . /go/src/github.com/luizalabs/teresa-api
 
 RUN make build-server
 
-CMD ["./teresa-server", "run", "--port", "8080"]
-EXPOSE 8080
+ENTRYPOINT ["./teresa-server"]
```
drgarcia1986: after this change, I always need to specify the `args` on k8s deploy yaml, right?
is it a good thing?

aguerra: Right. I guess, its considered a docker [best practice ](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#entrypoint), but the best thing is that you don't have to rebuild the image (you may have to pass a `--debug` flag in the future).

yagonobre: Maybe use `entrypoint` to binary and `cmd` to args

aguerra: @drgarcia1986 @yagonobre updated.

PR #250, opened 2017-08-09
build a k8s client per operation instead of using the same

drgarcia1986:
this change will help us with concurrency

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/250)


drgarcia1986:
thanks @aguerra , updated

PR #253, opened 2017-08-09
add helm chart

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/253)


helm/chart/teresa/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+        chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+        component: "server"
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: {{ template "name" . }}
+        image: "{{ .Values.docker.registry }}/{{ .Values.docker.image }}:{{ .Values.docker.tag }}"
+        {{- if .Values.tls.crt }}
+        args: ["run", "--tls"]
```
aguerra: Always appending `--tls` makes TLS stuff mandatory, am I right?

yagonobre: The `--tls` is append only if the `tls.crt` was defined

PR #255, opened 2017-08-09
organize pkg `client`

drgarcia1986:
Use root `cmd` pkg only for root command and `pkg/{server|client}/cmd` for regular commands with cobra stuff 🐍 .

This struct is heavy based on k8s project.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/255)


cmd/client/CHANGELOG.md
```diff
@@ -1,53 +0,0 @@
-# Change Log
```
yagonobre: Why remove changelog?

PR #257, opened 2017-08-10
add codecov integration

drgarcia1986:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/257)

### DIFF
--- a/.tests-run.sh
+++ b/.tests-run.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+rm -f coverage.txt
+
+for d in $(go list ./... | grep -v vendor); do
+    go test -coverprofile=profile.out $d
+    if [ -f profile.out ]; then
+        cat profile.out >> coverage.txt
+        rm profile.out
+    fi
+done
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,7 @@ go:
   - tip
 
 script:
-  - go test -v $(go list ./... | grep -v vendor)
+  - ./.tests-run.sh
+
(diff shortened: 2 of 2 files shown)


.tests-run.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+echo "" > coverage.txt
```
yagonobre: maybe `rm -f coverage.txt` looks better

arnaldopereira: does `coverage.txt` needs to be created before continuing?
if not, `rm -f` it. if so, either `echo "" > coverage.txt` or `rm -f coverage.txt; touch coverage.txt` would work.

.tests-run.sh
```diff
@@ -0,0 +1,12 @@
...
+
+for d in $(go list ./... | grep -v vendor); do
```
arnaldopereira: Maybe `grep -v ^vendor` to be specific

.tests-run.sh
drgarcia1986: let me check

.tests-run.sh
drgarcia1986: doesn't works because the package name is `github.com/luizalabs/teresa-api/vendor/github.com/juju`, i.e. both `go list` and `go tests` work with the package name not package dir.

.tests-run.sh
drgarcia1986: changed

drgarcia1986: works fine 👍

PR #261, opened 2017-08-11
only check password length in `set-password` command

drgarcia1986:
To prevent access troubles for users already saved with a weak password length.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/261)

### DIFF
--- a/pkg/client/cmd/user.go
+++ b/pkg/client/cmd/user.go
@@ -47,6 +47,9 @@ func setPassword(cmd *cobra.Command, args []string) {
 	if err != nil {
 		client.PrintErrorAndExit("Error trying to get the user password: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/client/cmd/user.go
```diff
@@ -47,6 +47,9 @@ func setPassword(cmd *cobra.Command, args []string) {
 	if err != nil {
 		client.PrintErrorAndExit("Error trying to get the user password: %v", err)
 	}
+	if err = client.CheckPasswordLength(p); err != nil {
```
arnaldopereira: What do you think of using `EnsurePasswordLength` instead? 💅

drgarcia1986: perfect 👍

drgarcia1986:
updated @arnaldopereira

PR #264, opened 2017-08-11
add storage type and s3 endpoint on helm chart

drgarcia1986:
Specific stuff to works fine with minio as slug storage.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa-api/264)

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -73,7 +73,10 @@ Parameter | Description | Default
 `db.hostname`| (Optional) Database hostname, if defined use mysql instead of sqlite| `""`
 `db.username` | (Optional) Database username | `""`
 `db.password` | (Optional) Database password | `""`
+`storage.type` | Type of storage | `s3`
+`aws.s3.force_path_style` | To force path style instead of subdomain-style | `false`
 `aws.s3.bucket` | S3 bucket path | `""`
+`aws.s3.endpoint` | (Optional) AWS Endpoint | `""`
 `aws.region` | AWS Region | `us-east-1`
 `aws.key.access` | AWS Access Key | `""`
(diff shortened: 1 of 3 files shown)


helm/README.md
```diff
@@ -73,7 +73,10 @@ Parameter | Description | Default
 `db.hostname`| (Optional) Database hostname, if defined use mysql instead of sqlite| `""`
 `db.username` | (Optional) Database username | `""`
 `db.password` | (Optional) Database password | `""`
+`storage.type` | Type of storage | `s3`
+`aws.s3.force_path_style` | To force path style instead of subdomain-style | `false`
 `aws.s3.bucket` | S3 bucket path | `""`
+`aws.s3.endpoint` | AWS Endpoint | `""`
```
yagonobre: I think that it's optional

helm/chart/teresa/templates/deployment.yaml
```diff
@@ -44,8 +44,14 @@ spec:
           value: /etc/teresa-keys/teresa.rsa
         - name: TERESA_SECRETS_PUBLIC_KEY
           value: /etc/teresa-keys/teresa.rsa.pub
+        - name: TERESA_STORAGE_TYPE
+          value: {{ .Values.storage.type }}
+        - name: TERESA_STORAGE_AWS_ENDPOINT
```
yagonobre: only append this if the `.Values.aws.s3.endpoint` was provide

helm/README.md
drgarcia1986: you're right!

drgarcia1986:
updated @yagonobre !
could you check again?

PR #278, opened 2017-08-15
feat(travis) update helm repo on teresa releases

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/278)


.helm.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+function download {
+	curl https://storage.googleapis.com/kubernetes-helm/helm-v2.5.1-linux-amd64.tar.gz --output /tmp/helm.tar.gz --silent
+
+	tar -x -C /tmp -f /tmp/helm.tar.gz linux-amd64/helm
+
+	if [[ $TRAVIS_TAG ]]; then
```
aguerra: Missing -z or -n right?

drgarcia1986: `-n` maybe ?

yagonobre: Maybe, but it was working in the two case. Install `aswcli` only if is a tag push

drgarcia1986: @yagonobre what do you think about using `-n` here? (to standardize with [here](https://github.com/luizalabs/teresa/pull/278/files#diff-90f57a2a951e5aa4477cc8c5e3b79904R34) )

yagonobre: updated

PR #280, opened 2017-08-17
Container specific limits for build and release phase

drgarcia1986:
According to k8s docs, the namespace limits works as default limit. So we can change the limits quota for any container that we create.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/280)


pkg/server/deploy/build.go
```diff
@@ -127,5 +136,9 @@ func newRunCommandSpec(a *app.App, deployId, command, slugURL string, fileStorag
 		fileStorage,
 	)
 	ps.Args = []string{"start", command}
+	ps.ContainerLimits = &ContainerLimits{
+		CPU:    opts.BuildLimitCPU,
+		Memory: opts.BuildLimitMemory,
+	}
 	return ps
```
drgarcia1986: I supposed yes, What do you think about it?

aguerra: So all apps will end up with 800m and 1Gi?

drgarcia1986: Again, I supposed yes 😂 
Could you explain how you thought this feature?

drgarcia1986: BTW, Both `800m` and `1Gi` are totally random values.

drgarcia1986: Ups, all apps no, all build and release phase.

aguerra: My bad, the patch formatting confused me.

drgarcia1986:
@yagonobre and @aguerra thanks for your review, I'll close this PR to do some toil work.

PR #281, opened 2017-08-17
Implement app autoscale

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/281)


pkg/client/cmd/app.go
```diff
@@ -423,6 +423,71 @@ WARNING:
...
+		Max: max,
+	}
+	req := &appb.SetAutoScaleRequest{
+		Name:      name,
+		AutoScale: as,
+	}
+	cli := appb.NewAppClient(conn)
+	if _, err := cli.SetAutoScale(context.Background(), req); err != nil {
+		fmt.Fprintln(os.Stderr, client.GetErrorMsg(err))
+		return
```
drgarcia1986: Use `PrintErrorAndExit`:
```go
client.PrintErrorAndExit(client.GetErrorMsg(err))
```

pkg/client/cmd/app.go
```diff
@@ -423,6 +423,71 @@ WARNING:
...
+	}
+	fmt.Println("Autoscale updated with success")
```
drgarcia1986: It will be nice if you print that message in green like the success login message.

pkg/protobuf/app/app.proto
```diff
@@ -9,6 +9,7 @@ service App {
     rpc SetEnv(SetEnvRequest) returns (Empty);
     rpc UnsetEnv(UnsetEnvRequest) returns (Empty);
     rpc List(Empty) returns (ListResponse);
+	rpc SetAutoScale(SetAutoScaleRequest) returns (Empty);
```
drgarcia1986: Is there an extra space (or tab)? 🤔

pkg/protobuf/app/app.proto
```diff
@@ -119,4 +120,15 @@ message UnsetEnvRequest {
     repeated string env_vars = 2;
 }
 
+message SetAutoScaleRequest {
+    string name = 1;
+	
+	message AutoScale {
+        int32 cpu_target_utilization = 1;
```
drgarcia1986: Why are you using `cpu_target_utilization` on protobuf specification if you don't currently use?

pkg/server/app/app.go
```diff
@@ -363,6 +364,30 @@ func (ops *AppOperations) List(user *database.User) ([]*AppListItem, error) {
...
 
+func (ops *AppOperations) SetAutoScale(user *database.User, appName string, as *AutoScale) error {
+	app, err := ops.checkPermAndGet(user, appName)
+	if err != nil {
+		return err
+	}
+
+	if err := ops.saveApp(app, user.Email); err != nil {
+		return teresa_errors.NewInternalServerError(err)
+	}
```
drgarcia1986: Why are you calling method `saveApp` ?

pkg/client/cmd/app.go
```diff
@@ -423,6 +423,71 @@ WARNING:
...
+		if max < 1 {
+			client.PrintErrorAndExit(fmt.Sprintf("--max=MAXPODS must be at least 1, max: %d", max))
+		}
+
+		if min != -1 && min > max {
+			client.PrintErrorAndExit(fmt.Sprintf("--max=MAXPODS must be larger or equal to --min=MINPODS, max: %d, min: %d", max, min))
+		}
+	} else if min == -1 {
+		client.PrintErrorAndExit("at least 1 flag is required")
+	}
```
drgarcia1986: What you think about to move this logic to another method and create some tests for it?

pkg/client/cmd/app.go
```diff
@@ -423,6 +423,71 @@ WARNING:
...
+		client.PrintErrorAndExit("invalid max parameter")
+	}
+
+	if max != -1 {
+		if max < 1 {
+			client.PrintErrorAndExit(fmt.Sprintf("--max=MAXPODS must be at least 1, max: %d", max))
+		}
+
+		if min != -1 && min > max {
+			client.PrintErrorAndExit(fmt.Sprintf("--max=MAXPODS must be larger or equal to --min=MINPODS, max: %d, min: %d", max, min))
```
drgarcia1986: Take a look at `PrintErrorAndExit` signature, that already expected a "format" and a slice of args, so you don't need to use `fmt.Sprintf` in this case.

yagonobre: Nice!

pkg/server/app/app.go
yagonobre: To update the last user that edit the app

pkg/protobuf/app/app.proto
yagonobre: I'm in doubt if it is time to implement it

pkg/protobuf/app/app.proto
yagonobre: my bad 😅

pkg/client/cmd/app.go
yagonobre: I think that the others command in app does not do it, what you think about update all?

pkg/server/app/app.go
drgarcia1986: Ok, LGTM but sounds a bit weird, I understand that it's necessary but still weird (change namespace annotation and autoScale).

I'm waiting for @aguerra opinion.

pkg/protobuf/app/app.proto
drgarcia1986: I avoid premature features, but you can follow your heart.

pkg/server/app/app.go
yagonobre: I think that `saveApp` only change notations on namespace

pkg/client/cmd/app.go
yagonobre: I thought about it, but  all command do input validation on his function, to me it looks simples, what you think?

pkg/client/cmd/app.go
drgarcia1986: Hum, maybe in another PR.

pkg/client/cmd/app.go
drgarcia1986: Take a look at deploy command, there are some separate funcs.

pkg/server/app/models.go
```diff
@@ -262,3 +262,30 @@ func newListResponse(items []*AppListItem) *appb.ListResponse {
...
 }
+
+func newAutoScale(req *appb.SetAutoScaleRequest) *AutoScale {
+	return &AutoScale{
+		Max: req.AutoScale.Max,
+		Min: req.AutoScale.Min,
+	}
+}
+
+func updateAutoScale(app *App, old, new *AutoScale) {
```
drgarcia1986: avoid reserved words like `new`

pkg/client/cmd/app.go
```diff
@@ -423,6 +423,70 @@ WARNING:
 	Run: appLogs,
 }
 
+var appAutoScaleSetCmd = &cobra.Command{
+	Use:   "autoscale <name> [flags]",
+	Short: "Set autoscale parameters for the app",
+	Long: `Set auto scaling parameters for the application.
+
+You can configure the autoscale parameters from the application. (i.e. the minimum and maximum number of replicas)
```
aguerra: I guess `for the application` sounds better.

pkg/client/cmd/app.go
```diff
@@ -423,6 +423,70 @@ WARNING:
...
+		if max < 1 {
+			client.PrintErrorAndExit("--max=MAXPODS must be at least 1, max: %d", max)
+		}
+
+		if min != -1 && min > max {
+			client.PrintErrorAndExit("--max=MAXPODS must be larger or equal to --min=MINPODS, max: %d, min: %d", max, min)
+		}
+	} else if min == -1 {
+		client.PrintErrorAndExit("at least 1 flag is required")
+	}
```
aguerra: What do you guys think of testing max and min independently? If I got it right, passing only `--min=-42` will work on the client

aguerra: @drgarcia1986

pkg/server/app/app.go
aguerra: Putting `saveApp` in the end makes it a lot less weird for me as all worked and we are now just updating the last user. For the future a new method sounds better.

pkg/server/app/models.go
```diff
@@ -262,3 +262,30 @@ func newListResponse(items []*AppListItem) *appb.ListResponse {
...
+			if new.Min <= old.Max {
+				new.Max = old.Max
+			} else {
+				new.Max = new.Min
+			}
+		}
+	}
+	new.CPUTargetUtilization = old.CPUTargetUtilization
+	app.AutoScale = new
+}
```
aguerra: What do you guys think of always requiring the min and max parameters? It seems it will simplify this logic a lot...

aguerra: I mean on the client...

pkg/client/cmd/app.go
yagonobre: Yes, you can pass only one flag or both

pkg/server/app/models.go
yagonobre: yes, it simplifies in client and server, I try to follow the in the implementation of `kubectl autoscale`, but in `kubectl` the max is ever required.

drgarcia1986: I agree with @aguerra on this point.

pkg/client/cmd/app.go
```diff
@@ -423,6 +425,90 @@ WARNING:
...
+	if err != nil {
+		client.PrintErrorAndExit("invalid max parameter")
+	}
+
+	cpu, err := cmd.Flags().GetInt32("cpu-percent")
+	if err != nil {
+		client.PrintErrorAndExit("invalid cpu-percent parameter")
+	}
+
+	if msg, isValid := validateFlags(min, max); isValid != true {
```
drgarcia1986:
```diff
- if msg, isValid := validateFlags(min, max); isValid != true {
+ if msg, isValid := validateFlags(min, max); !isValid  {
```

pkg/client/cmd/app.go
```diff
@@ -423,6 +425,90 @@ WARNING:
...
+		AutoScale: as,
+	}
+	cli := appb.NewAppClient(conn)
+	if _, err := cli.SetAutoScale(context.Background(), req); err != nil {
+		client.PrintErrorAndExit(client.GetErrorMsg(err))
+	}
+	fmt.Println("Autoscale updated with success")
+}
+
+func validateFlags(min, max int32) (msg string, isValid bool) {
```
drgarcia1986: What do you think about don't use named return values?
I'm thinking about this func and sounds like unnecessary.
It's simple to write something like this:
```go
if max == flagNotDefined || flagNotDefined == {
    return "--min and --max are required`, false
}
```
With this change you don't need to use `else` f.ex/

pkg/client/cmd/app.go
```diff
@@ -452,6 +539,10 @@ func init() {
 	// App logs
 	appLogsCmd.Flags().Int64("lines", 10, "number of lines")
 	appLogsCmd.Flags().Bool("follow", false, "follow logs")
+	// App autoscale
+	appAutoScaleSetCmd.Flags().Int32("min", -1, "Auto scale min size.")
```
drgarcia1986: you can use `flagNotDefined` here too.

pkg/client/cmd/app_test.go
```diff
@@ -0,0 +1,25 @@
+package cmd
+
+import "testing"
+
+var flagTests = []struct {
+	min    int32
+	max    int32
+	output bool
```
drgarcia1986: We were using `expected` for those cases.

pkg/client/cmd/app_test.go
```diff
@@ -0,0 +1,25 @@
...
+	{5, -10, false},
+	{-10, 5, false},
+	{-10, -10, false},
+}
+
+func TestValidateFlags(t *testing.T) {
+	for _, tt := range flagTests {
+		_, isValid := validateFlags(tt.min, tt.max)
+		if isValid != tt.output {
+			t.Errorf("Validate fail, expect %t and get %t (min = %d, max = %d).", tt.output, isValid, tt.min, tt.max)
```
drgarcia1986: `got` sounds better, what do you think?

pkg/client/cmd/app.go
yagonobre: I think that named return in this function is cleaner but ok.

yagonobre:
@aguerra @drgarcia1986 updated

pkg/server/k8s/client.go
```diff
@@ -259,15 +259,18 @@ func (k *k8sClient) CreateSecret(appName, secretName string, data map[string][]b
...
 		return err
 	}
 
 	hpa := newHPA(a)
 
-	_, err = kc.AutoscalingV1().HorizontalPodAutoscalers(a.Name).Create(hpa)
+	_, err = kc.AutoscalingV1().HorizontalPodAutoscalers(a.Name).Update(hpa)
+	if k.IsNotFound(err) {
+		_, err = kc.AutoscalingV1().HorizontalPodAutoscalers(a.Name).Create(hpa)
+	}
```
drgarcia1986: 💅 what do you think about using an "object" like `hpa`:
```golang
hpa := kc.AutoscalingV1().HorizontalPodAutoscalers(a.name)
...
hpa.Update(hpaSpec)
```

yagonobre: I think that it is not the pattern used

pkg/client/cmd/app.go
```diff
@@ -429,6 +432,85 @@ WARNING:
 	Run: appLogs,
 }
 
+var appAutoScaleSetCmd = &cobra.Command{
+	Use:   "autoscale <name> [flags]",
+	Short: "Set autoscale parameters for the app",
+	Long: `Set auto scaling parameters for the application.
+
+You can configure the autoscale parameters for the application. (i.e. the minimum and maximum number of replicas)
```
arnaldopereira: That help text might be improved, suggestion:

    Set application's autoscaling.

    You can set the lower and upper limit for the number of pods of the application, as well as the
    target CPU utilization to trigger the autoscaler.

pkg/client/cmd/app.go
```diff
@@ -458,6 +541,10 @@ func init() {
 	// App logs
 	appLogsCmd.Flags().Int64("lines", 10, "number of lines")
 	appLogsCmd.Flags().Bool("follow", false, "follow logs")
+	// App autoscale
+	appAutoScaleSetCmd.Flags().Int32("min", flagNotDefined, "Auto scale min size.")
```
arnaldopereira: `Minimium number of replicas`

pkg/client/cmd/app.go
```diff
@@ -458,6 +541,10 @@ func init() {
...
+	appAutoScaleSetCmd.Flags().Int32("max", flagNotDefined, "Auto scale max size.")
```
arnaldopereira: `Maximum number of replicas`

pkg/client/cmd/app.go
```diff
@@ -458,6 +541,10 @@ func init() {
...
+	appAutoScaleSetCmd.Flags().Int32("cpu-percent", flagNotDefined, "The target average CPU utilization (represented as a percent of requested CPU) over all the pods. If it's not specified or negative, the current autoscaling policy will be used.")
```
arnaldopereira: `If it's not specified or negative, the default kubernetes autoscaling policy will be used`

arnaldopereira: `If it's not specified or negative, the default kubernetes autoscaling policy will be used`

pkg/protobuf/app/app.proto
```diff
@@ -119,4 +120,15 @@ message UnsetEnvRequest {
...
 }
 
+message SetAutoScaleRequest {
+    string name = 1;
+
+	message AutoScale {
+        int32 cpu_target_utilization = 1;
+		int32 max = 2;
+		int32 min = 3;
+    }
```
arnaldopereira: Wrong formatting.

yagonobre: 👀

pkg/server/app/app.go
```diff
@@ -363,6 +364,33 @@ func (ops *AppOperations) List(user *database.User) ([]*AppListItem, error) {
 	return items, nil
 }
 
+func (ops *AppOperations) SetAutoScale(user *database.User, appName string, as *AutoScale) error {
```
arnaldopereira: On `SetAutoScale`, S should be lower case. In general `autoscale` should be treated as a single word - this applies to all the occurrences throughout this PR.

pkg/server/app/app_test.go
```diff
@@ -704,3 +704,57 @@ func TestAppOperationsUnsetEnvErrInternalServerErrorOnSaveApp(t *testing.T) {
 		t.Errorf("expected ErrInternalServerError, got %v", err)
 	}
 }
+
+func TestAppOperationsSetAutoScale(t *testing.T) {
```
arnaldopereira: Should we really use `AppOperations` on every test function?

pkg/client/cmd/app.go
yagonobre: `defaults` seems that the same value will be used to all app

pkg/server/app/app_test.go
yagonobre: It may be good in case of test fail

pkg/server/app/app.go
yagonobre: nice :)

PR #282, opened 2017-08-17
limit the size of app name

yagonobre:
This is for avoiding problem with kube-dns.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/282)

### DIFF
--- a/pkg/client/cmd/app.go
+++ b/pkg/client/cmd/app.go
@@ -17,6 +17,9 @@ import (
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/client/cmd/app.go
```diff
@@ -56,6 +56,9 @@ func createApp(cmd *cobra.Command, args []string) {
 		return
 	}
 	name := args[0]
+	if len(name) > 63 {
```
drgarcia1986: Try to avoid magic numbers, at least put in a constant with a descriptive name is a good idea.

PR #284, opened 2017-08-18
remove `max-cpu` and `max-memory` flags of app create command

drgarcia1986:
What do you think? Is it a good idea to remove those values from the server too?

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/284)


yagonobre:
I think that is nice remove this in the server too

drgarcia1986:
@yagonobre great, I'll do in another PR, ok? (because isn't mandatory for a new release f.ex)

drgarcia1986:
(I) we change our(my) mind.

PR #285, opened 2017-08-18
Add support for Teresa yaml files per process type

aguerra:
By default teresa-processtype.yaml is used with teresa.yaml
as a fallback.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/285)


pkg/server/deploy/config.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 const (
+	ProcfileFileName       = "Procfile"
+	teresaYamlFileNameTmpl = "teresa%s%s.yaml"
```
yagonobre: It's not more exported?

yagonobre: Is it not more exported?

yagonobre: Nice :)

pkg/server/deploy/config.go
```diff
@@ -85,28 +92,26 @@ func getDeployConfigFilesFromTarBall(tarBall io.ReadSeeker) (*DeployConfigFiles,
...
+				if err := deployFiles.fillTeresaYaml(tarReader); err != nil {
+					return nil, err
+				}
 			}
 		} else {
 			deployFiles.Procfile = make(map[string]string)
 			if err := readFileFromTarBall(tarReader, deployFiles.Procfile); err != nil {
 				return nil, err
 			}
 		}
```
drgarcia1986: Why you removed that last if? (`if deployFiles.TeresaYaml != nil && deployFiles.Procfile != nil`).
That "if" will prevent Teresa to read a bunch of files unnecessary, or am I miss something?

aguerra: Suppose that `deployFiles.TeresaYaml` is filled (not null) with `teresa.yaml`, `teresa-processtype.yaml` maybe the last one when iterating, so we can't break the loop. What do you think?

drgarcia1986: You have a point!

PR #292, opened 2017-08-23
Fix nil pointer dereference on app info command

aguerra:
On the presence of pending pods for instance.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/292)


aguerra:
As this line was added for this release, we decided to skip the changelog.

PR #297, opened 2017-08-23
Add --debug flag to the server

aguerra:
For now only prints the stack trace on every panic/recover.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/297)


pkg/server/cmd/run.go
```diff
@@ -75,6 +76,8 @@ func runServer(cmd *cobra.Command, args []string) {
 		log.Fatal("Error getting deploy configuration:", err)
 	}
 
+	debug := len(os.Getenv("TERESA_SERVER_DEBUG")) > 0
```
yagonobre: Maybe use `envconfig`?

aguerra:
envconfig won!

aguerra:
@drgarcia1986 @yagonobre updated.

PR #307, opened 2017-08-24
Fix hanging deploys when the app dir doesn't exist

aguerra:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/307)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - [server] Restart count on 'app info' output
 - [client] Validate 'deploy' command parameters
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - [server] Restart count on 'app info' output
+- [client] Hanging deploys when the app dir doesn't exist
```
arnaldopereira: This actually **fixes** hanging deploys, right?

aguerra: Yes, typing in a hurry...

drgarcia1986: IMO That's right because that description is already in `fixed` section of the changelog.

PR #314, opened 2017-08-29
[HELM] use minio on teresa chart

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/314)

### DIFF
--- a/helm/chart/teresa/values.yaml
+++ b/helm/chart/teresa/values.yaml
@@ -30,3 +30,10 @@ build:
     cpu: 500m
     memory: 1024Mi
 debug: false
+useMinio: false
+minio:
+  serviceType: ClusterIP
+  bucketDefault:
+    enabled: true
+    name: teresa
+    policy: public
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ ENV/
 # Rope project settings
 .ropeproject
 *.sqlite
+
+#Helm Dependency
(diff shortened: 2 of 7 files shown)


helm/chart/teresa/values.yaml
```diff
@@ -30,3 +30,6 @@ build:
     cpu: 500m
     memory: 1024Mi
 debug: false
+useMinio: false
+minio:
+  serviceType: ClusterIP
```
drgarcia1986: why not ELB? are you thinking about minikube?

yagonobre: I guess that the `minio` will only be used by `teresa` then I don't need an ELB. Makes sense?

drgarcia1986: I'm not sure about that. (Don't forget you need to create the bucket before use with Teresa)

yagonobre: Now this create the bucket

yagonobre:
waiting https://github.com/kubernetes/charts/pull/2061

drgarcia1986:
What about this PR ?

yagonobre:
@drgarcia1986 needs rebase but LGTM

yagonobre:
@drgarcia1986 done

yagonobre:
@aguerra @drgarcia1986 rebased

PR #316, opened 2017-08-30
run travis only in pr and tag push

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/316)


yagonobre:
```yaml
on:
  tags: true
```
work only in deployment scope

PR #320, opened 2017-08-31
don't print pod without status on app info

yagonobre:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/320)


yagonobre:
Done :)

PR #321, opened 2017-08-31
Implement 'team remove-user' command

aguerra:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/321)


pkg/client/cmd/team.go
```diff
@@ -45,12 +45,25 @@ You need to create a user before use this command.`,
...
 }
 
+var teamRemoveUserCmd = &cobra.Command{
+	Use:   "remove-user",
+	Short: "Remove a member of a team",
+	Long: `Remove a member of a team.
+
+You can remove an user of a team with:
+
+  $ teresa team remove-user --user john.doe@foodomain.com --team foo
```
yagonobre: Maybe user can be an arg instead of a flag

yagonobre: Maybe user can be an arg instead of a flag

pkg/protobuf/team/team.proto
```diff
@@ -19,6 +20,11 @@ message AddUserRequest {
     string user = 2;
 }
 
+message RemoveUserRequest {
+    string name = 1;
```
yagonobre: Maybe `team_name`

pkg/client/cmd/team.go
aguerra: I think that for now we need consistency with `add-user` but it is an interesting option for another PR.

yagonobre: Nice :)

PR #325, opened 2017-09-08
App Delete Implementation

lucianokadoya-luizalabs:
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/325)


pkg/client/cmd/app.go
```diff
@@ -558,6 +610,8 @@ func init() {
 	// App unset env vars
 	appEnvUnSetCmd.Flags().String("app", "", "app name")
 	appEnvUnSetCmd.Flags().Bool("no-input", false, "unset env vars without warning")
+	// App Delete
+	appDelCmd.Flags().String("app", "", "app name")
```
yagonobre: You no need this flag

pkg/client/cmd/app.go
```diff
@@ -205,6 +205,57 @@ func appList(cmd *cobra.Command, args []string) {
...
+	Example: "  $ teresa app delete foo",
+	Run:     appDel,
+}
+
+func appDel(cmd *cobra.Command, args []string) {
+	if len(args) == 0 || len(args) > 1 {
+		cmd.Usage()
+		return
+	}
+
```
yagonobre: Maybe remove this line

pkg/client/cmd/app.go
```diff
@@ -205,6 +205,57 @@ func appList(cmd *cobra.Command, args []string) {
...
+var appDelCmd = &cobra.Command{
+	Use:     "delete <name>",
+	Short:   "Delete app",
+	Long:    "Delete app.",
+	Example: "  $ teresa app delete foo",
+	Run:     appDel,
+}
+
+func appDel(cmd *cobra.Command, args []string) {
+	if len(args) == 0 || len(args) > 1 {
```
drgarcia1986: What do you think about `if len(args) != 1 {` ?

pkg/client/cmd/app.go
```diff
@@ -205,6 +205,57 @@ func appList(cmd *cobra.Command, args []string) {
...
+		return
+	}
+
+	fmt.Print("Please re type the app name: ")
+	resp, _ := bufio.NewReader(os.Stdin).ReadString('\n')
+	resp = strings.ToLower(strings.TrimRight(resp, "\r\n"))
+	if resp != name {
+		fmt.Println("Delete process aborted!")
+		return
+	} else {
```
drgarcia1986: you don't need that `else` 'cause you already exit function one line before, f.ex:
```go
if resp != name {
    fmt.Println("Delete process aborted!")
    return
}
_, err := cli.Delete(context.Background(), &appb.DeleteRequest{Name: name})
...
```

pkg/client/cmd/app.go
```diff
@@ -205,6 +205,57 @@ func appList(cmd *cobra.Command, args []string) {
...
+	s, _ := bufio.NewReader(os.Stdin).ReadString('\n')
+	s = strings.ToLower(strings.TrimRight(s, "\r\n"))
+	if s != "yes" {
+		fmt.Println("Delete process aborted!")
+		return
+	}
+
+	fmt.Print("Please re type the app name: ")
+	resp, _ := bufio.NewReader(os.Stdin).ReadString('\n')
+	resp = strings.ToLower(strings.TrimRight(resp, "\r\n"))
```
drgarcia1986: What do you think about "DRY" this routine? f.ex:
```go
func getStdin(msg string) (string, error) {
    fmt.Print(msg)
    resp, err := bufio.NewReader(os.Stdin).ReadString('\n')
    if err != nil {
        return "", err
    }
    return strings.ToLower(strings.TrimRight(resp, "\r\n")), nil
}
```
If you agree, I sugest you tu put this func on `pkg/client/[something]` instead of directly on `cmd` pkg.

pkg/server/app/fake_test.go
```diff
@@ -317,3 +317,34 @@ func TestFakeOperationsSetAutoscaletErrNotFound(t *testing.T) {
...
 }
+
+func TestFakeOperationsDelete(t *testing.T) {
+	fake := NewFakeOperations()
+	user := &database.User{Name: "gopher@luizalabs.com"}
+	app := &App{Name: "teresa"}
+	fake.(*FakeOperations).Storage[app.Name] = app
+
+	if err := fake.Delete(user, app.Name); err != nil {
+		t.Fatal("error on Delete: ", err)
```
drgarcia1986: for the last test in a test function is ok to use `t.Error`, we use `t.Fatal` only if we want to stop the test.

drgarcia1986:
There're some conflicts with the master, could you rebase with the master?

pkg/client/cmd/app.go
lucianokadoya-luizalabs: Ok. I agree @drgarcia1986 . I utilized this new package to use in `appEnvSet` and `appEnvUnSet` function too.

pkg/client/getstdin/getstdin.go
```diff
@@ -0,0 +1,17 @@
+package getstdin
```
drgarcia1986: sorry to be an annoying guy but, you don't need to create a new package, just `client.GetStdin()` looks good to me. (if you move this file to `pkg/client/getstdin.go` and change the package name to `client` will be great)

pkg/client/cmd/app.go
```diff
@@ -205,6 +205,52 @@ func appList(cmd *cobra.Command, args []string) {
...
+	}
+
+	fmt.Print()
+	s, _ := getstdin.GetStdin("Are you sure? (yes/NO)? ")
+	if s != "yes" {
+		fmt.Println("Delete process aborted!")
+		return
+	}
+
+	resp, _ := getstdin.GetStdin("Please re type the app name: ")
```
aguerra: typo: `retype`

pkg/server/app/fake.go
```diff
@@ -94,6 +94,21 @@ func (f *FakeOperations) List(user *database.User) ([]*AppListItem, error) {
...
+	if !hasPerm(user.Email) {
+		return auth.ErrPermissionDenied
+	}
+
+	if _, found := f.Storage[appName]; !found {
+		return ErrNotFound
+	}
+
+	return nil
+}
```
aguerra: You should delete the key here,

pkg/client/getstdin/getstdin.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func GetStdin(msg string) (string, error) {
+	fmt.Print(msg)
+	resp, err := bufio.NewReader(os.Stdin).ReadString('\n')
+	if err != nil {
+		return "", err
+	}
+	return strings.ToLower(strings.TrimRight(resp, "\r\n")), nil
+}
```
aguerra: Maybe `getstdin` as name is too specific.

pkg/client/cmd/app.go
```diff
@@ -410,9 +454,7 @@ func appEnvUnset(cmd *cobra.Command, args []string) {
 		client.PrintErrorAndExit("Invalid no-input parameter")
 	}
 	if !noinput {
-		fmt.Print("Are you sure? (yes/NO)? ")
-		s, _ := bufio.NewReader(os.Stdin).ReadString('\n')
-		s = strings.ToLower(strings.TrimRight(s, "\r\n"))
+		s, _ := getstdin.GetStdin("Are you sure? (yes/NO)? ")
```
aguerra: Need to check `getstdin.GetStdin` errors.

pkg/client/getstdin/getstdin.go
lucianokadoya-luizalabs: no problem man.

pkg/server/app/fake.go
drgarcia1986: deleted

PR #337, opened 2017-09-12
sort envs in app info

### LINKED ISSUE #336: Sort env vars.
It would be nice if `teresa app info myapp` return env vars sorted.

yagonobre:
fixes #336

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/337)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Deploys are now performed using the 'deploy create' command
+- app info now print env vars sorted
 
 ### Fixed
 - App info don't print pod without status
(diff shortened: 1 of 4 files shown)


pkg/server/app/sort.go
```diff
@@ -0,0 +1,21 @@
+package app
```
drgarcia1986: why not in `pkg/server/client` ?

pkg/server/app/sort.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "sort"
+
+type ByKey []*EnvVar
```
drgarcia1986: maybe `EnvVarsByKey`?

yagonobre: I think that `ByKey` is more cleaner, like in sort documentation

drgarcia1986: but think about usave 
```go
import "pkg/client/app/sort"
...
sort.ByKey
```
`ByKey` of what?

yagonobre: you can use the func `sortEnvsByKey`

PR #348, opened 2017-10-04
check --cluster flag when saving token to config file in teresa login 

yagonobre:
The `teresa login` ignore the `--cluster` flag on saving token to config file, this pr fix it.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/348)


drgarcia1986:
please add an entry on changelog (on `fix` section)

PR #356, opened 2017-10-09
configurable timeout of `PodRun` process

drgarcia1986:
A bunch migrations are critical and (unfortunately) makes locks on the database. So, to prevent any dangerous behaviors, we can configure the timeout of pod run process to a large duration.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Upgrade `slugbuilder` version to `v3.0.0`
+- Timeout of `PodRun` process (deploy and release) now is configurable
```
arnaldopereira: `is now configurable`

PR #370, opened 2017-11-09
add current env to app info

gerep:
It is possible for users to manipulate an APP in the wrong environment
and to avoid that I suggest adding the current environment when listing
the APP info.

A possible downside of this change is the fact that it is using the
function `ReadConfigFile (cfgFile)) (*Config, error)` to get the current
cluster name.

---
This change is [Reviewable](https://reviewable.io/reviews/luizalabs/teresa/370)

### DIFF
--- a/pkg/client/cmd/app.go
+++ b/pkg/client/cmd/app.go
@@ -280,6 +280,13 @@ func appInfo(cmd *cobra.Command, args []string) {
 	color.New(color.FgCyan, color.Bold).Printf("[%s]\n", name)
 	bold := color.New(color.Bold).SprintFunc()
 
+	f, err := client.ReadConfigFile(cfgFile)
+	if err != nil {
+		client.PrintErrorAndExit("Cannot read the config file, have you created it with `teresa set-cluster` command?")
+	}
+
+	color.New(color.FgRed, color.Bold).Printf("env: %s\n", f.CurrentCluster)
+
 	fmt.Println(bold("team:"), info.Team)
 	if len(info.Addresses) > 0 {
 		fmt.Println(bold("addresses:"))


drgarcia1986:
I'm not sure 🤔 
Maybe use the `--cluster` flag is a better idea (to avoid confusions like that).
Or, a global flag to show the current cluster in each command (is it possible? IDK 😐 )

Hey @yagonobre and @aguerra , what you think about it?

pkg/client/cmd/app.go
```diff
@@ -280,6 +280,13 @@ func appInfo(cmd *cobra.Command, args []string) {
...
 	bold := color.New(color.Bold).SprintFunc()
 
+	f, err := client.ReadConfigFile(cfgFile)
+	if err != nil {
+		client.PrintErrorAndExit("Cannot read the config file, have you created it with `teresa set-cluster` command?")
+	}
+
+	color.New(color.FgRed, color.Bold).Printf("env: %s\n", f.CurrentCluster)
+
 	fmt.Println(bold("team:"), info.Team)
```
aguerra: I'm in favor of  showing the environment but this way you are reading the config file two times (the first one in `connection.New`).

yagonobre:
@gerep Maybe you can print the cluster name in your PS1, what you think?

gerep:
@yagonobre You mean by using `fmt.Println`?

I don't get what is the cost of showing it as suggested, together with the APP info, can you clarify?

@aguerra Indeed! This has to be improved and will if we agree with a solution =)

pkg/client/cmd/app.go
arnaldopereira: Maybe saving the `current` cluster in some variable to use afterwards.

Either case, this information would be a cluster nickname and not an environment necessarily.

aguerra:
Closing as we need some refactoring on the connection to properly implement it.

PR #390, opened 2018-01-11
Fix tar file creation on deploys

aguerra:
Do it before openning a connection to the server and remove the file before exiting.

### DIFF
--- a/pkg/client/cmd/deploy.go
+++ b/pkg/client/cmd/deploy.go
@@ -234,6 +234,9 @@ func deployApp(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	tarPath := createAppTarBall(appName, appFolder)
+	defer os.Remove(tarPath)
+
 	conn, err := connection.New(cfgFile, currentClusterName)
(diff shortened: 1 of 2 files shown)


pkg/client/cmd/deploy.go
```diff
@@ -297,7 +303,7 @@ func sendAppTarball(appName, appFolder string, stream dpb.Deploy_MakeClient) err
 			Chunk: buf,
 		}}}
 		if err := stream.Send(bufMsg); err != nil {
-			fmt.Fprintln(os.Stderr, "Error sending tarball chunk:")
+			fmt.Fprintln(os.Stderr, "Error sending tarball chunk: %v", err)
```
drgarcia1986: maybe`Fprintf` or `Fprintln(os.Stderr, "Error sending tarball chunk:", err)` ?

aguerra: yeah I'm going blind for sure...

drgarcia1986: 😂

PR #395, opened 2018-01-16
set password admin

yagonobre:


pkg/client/cmd/user.go
```diff
@@ -136,4 +146,5 @@ func init() {
 	deleteUserCmd.Flags().String("email", "", "user email [required]")
 
 	RootCmd.AddCommand(setUserPasswordCmd)
+	setUserPasswordCmd.Flags().String("user", "", "user to set the password, if not provide will set the current user password")
```
aguerra: provide -> provided

pkg/server/user/user.go
```diff
@@ -61,23 +61,32 @@ func (dbu *DatabaseOperations) GetUser(email string) (*database.User, error) {
...
 	if err != nil {
 		return err
 	}
+
 	pass, err := bcrypt.GenerateFromPassword([]byte(newPassword), bcrypt.DefaultCost)
 	if err != nil {
 		return teresa_errors.New(
 			teresa_errors.ErrInternalServerError,
-			errors.Wrap(err, fmt.Sprintf("Generating the password hash to user %s", email)),
+			errors.Wrap(err, fmt.Sprintf("Generating the password hash to user %s", u.Email)),
```
aguerra: I guess it's email on errors.Wrap as email may be different from u.Email, right?

pkg/server/user/user_test.go
```diff
@@ -111,28 +111,79 @@ func TestDatabaseOperationsSetPassword(t *testing.T) {
...
+	if err = dbu.SetPassword(user, expectedPassword, ""); err != nil {
 		t.Fatal("error trying to set a new password: ", err)
 	}
 	if _, err = dbu.Login(expectedEmail, expectedPassword, time.Second); err != nil {
 		t.Error("error trying to make login with new password: ", err)
 	}
 }
 
-func TestDatabaseOperationsSetPasswordForInvalidUser(t *testing.T) {
+func TestDatabaseOperationsSetPasswordAdmin(t *testing.T) {
```
aguerra: This Admin suffix is  a little bit misleading as there as tests with non admin users.

pkg/server/user/user.go
yagonobre: Yes, I'll fix this, thanks!

pkg/server/user/user_test.go
yagonobre: Yes, this is a mistake

yagonobre:
@aguerra updated

PR #398, opened 2018-01-22
Add deploy default lifecycle with 10s drain timeout

aguerra:


pkg/server/deploy/build_test.go
```diff
@@ -166,3 +166,25 @@ func TestNewDeploySpec(t *testing.T) {
...
+		expectedDescription,
+		expectedSlugURL,
+		expectedProcessType,
+		opts,
+	)
+
+	if ds.Lifecycle == nil {
+		t.Error("expected lifecycle; got nil")
+	}
+}
```
drgarcia1986: What do you think about performing the test against the default value instead of check for a no nil value?

aguerra: Sure, I also will change the test name to reflect better what is being tested,

aguerra:
@drgarcia1986 updated.

PR #401, opened 2018-01-24
Rewrite client tar pkg and change the deploy accordingly

aguerra:
- simpler code with no external deps
- no filesystem race conditions and chokes on symlinks
- tests


pkg/client/cmd/deploy.go
```diff
@@ -268,16 +204,8 @@ func deployApp(cmd *cobra.Command, args []string) {
...
-	fmt.Println("Generating tarball of:", appFolder)
-	tarPath, err := createTempArchiveToUpload(appName, appFolder)
-	if err != nil {
-		client.PrintErrorAndExit("Error generating tarball: %v", err)
-	}
-	return tarPath
-}
-
 func sendAppTarball(tarPath string, stream dpb.Deploy_MakeClient) error {
+	fmt.Println("Sending app tarbal...")
```
drgarcia1986: 👏

pkg/client/tar/tar.go
```diff
@@ -1,47 +1,93 @@
...
+	file, err := os.Open(path)
+	if err != nil {
+		return errors.Wrap(err, "failed to open file")
+	}
+	defer file.Close()
 
-func (t *TarFile) Close() error {
-	return t.writer.Close()
+	_, err = io.Copy(tw, file)
+	return errors.Wrap(err, "failed to copy file contents to tarball")
```
drgarcia1986: Always get an error? 
if not, what do you thinking about?:
```go
if _, err := io.Copy(tw, file); err != nil {
    return errors.Wrap(err, "failed to copy file contents to tarball")
}
return nil
```
I'm not sure if `errors.Wraps` returns `nil` for non errors args

aguerra: Yes, it does.

Issue #347, opened 2017-09-25
Shell


jairhenrique:
As a teresa user
I want to access the shell from a pod
To run cli commands from my app

drgarcia1986:
What about running a command without an interactive shell inside of a pod?

jairhenrique:
>What about running a command without an interactive shell inside of a pod?

I think it's good enough, for now!

drgarcia1986:
`teresa exec` command will be delivery in the next version, stay alert 😄

PR #402, opened 2018-01-31
Add `exec` command (to exec a command on a replica of current deployment)

drgarcia1986:
Also, reorganize `deploy` pkg.


drgarcia1986:
Related to #347

pkg/protobuf/exec/exec.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package exec;
+
+service Exec {
+    rpc Command(CommandRequest) returns (stream CommandResponse);
+}
+
+message CommandRequest {
+    string name = 1;
```
aguerra: maybe `app_name`, when I first saw I thought it was the command name

pkg/server/exec/exec.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/luizalabs/teresa/pkg/server/auth"
+	"github.com/luizalabs/teresa/pkg/server/database"
+	"github.com/luizalabs/teresa/pkg/server/spec"
+	"github.com/luizalabs/teresa/pkg/server/storage"
+)
+
+type Operations interface {
+	Command(user *database.User, appName string, command ...string) (io.ReadCloser, <-chan error)
+	CommandBySpec(podSpec *spec.Pod) (io.ReadCloser, <-chan error)
+}
```
aguerra: Maybe `RunCommand` and `RunPod` (`PodRun` ?)

pkg/server/exec/exec.go
```diff
@@ -0,0 +1,106 @@
...
+	errChan := make(chan error, 1)
+	a, err := ops.appOps.Get(appName)
+	if err != nil {
+		errChan <- err
+		return nil, errChan
+	}
+	if !ops.appOps.HasPermission(user, appName) {
+		errChan <- auth.ErrPermissionDenied
+		return nil, errChan
+	}
```
aguerra: Can `CheckPermAndGet` be used here?

pkg/server/exec/exec.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+		command...,
+	)
+
+	return ops.CommandBySpec(podSpec)
+}
+
+func (ops *ExecOperations) CommandBySpec(podSpec *spec.Pod) (io.ReadCloser, <-chan error) {
+	errChan := make(chan error)
+	r, w := io.Pipe()
```
aguerra: Shouldn't it be `make(chan error, 1)`. Otherwise `errChan <- err` will block, right?

pkg/server/exec/handlers.go
```diff
@@ -0,0 +1,49 @@
+package exec
+
+import (
+	"github.com/luizalabs/teresa/pkg/goutil"
+	execpb "github.com/luizalabs/teresa/pkg/protobuf/exec"
+	"github.com/luizalabs/teresa/pkg/server/database"
+	"google.golang.org/grpc"
+)
```
aguerra: Maybe skip a line to separate the google import

pkg/server/exec/handlers.go
```diff
@@ -0,0 +1,49 @@
...
+				return nil
+			}
+			msg = m
+		}
+
+		if err := stream.Send(&execpb.CommandResponse{Text: msg}); err != nil {
+			return err
+		}
+	}
+}
```
aguerra: Maybe add `ctx.Done()` but I don't know the complexity so follow your heart.

pkg/client/cmd/exec.go
```diff
@@ -0,0 +1,65 @@
...
+var execCmd = &cobra.Command{
+	Use:   "exec",
+	Short: "Exec a command on an app replica",
+	Long: `Exec a command on an app replica.
+
+You can execute a command on an app replica (same of current deploy),
+Teresa will collect and stream the stdout of replica until the command ends`,
+	Example: ` $ teresa exec <app-name> -- python manage.py start_job_x -a arg1 -s arg2`,
+	Run:     execCommand,
+}
```
aguerra: Maybe it would be wise to say non-interactive command (shell session won't work for example). Also it would be nice to indent the example based on the other commands.

pkg/server/exec/handlers.go
drgarcia1986: I'm using`goimports` 🤔

pkg/server/exec/exec.go
drgarcia1986: I was thinking in the complete command call (with the pkg), e.g. `exec.Command` (instead of `exec.RunCommand`), but yes, I have some doubts... `exec.RunCommand` sounds good to me.

pkg/server/exec/handlers.go
drgarcia1986: Good point, I'll move the logic of `ctx.Done()` + `cancelation` of pkg deploy to pkg exec, what do you think about it?

pkg/server/exec/exec.go
drgarcia1986: Yes, but I'm writing in this channel inside a goroutine, I think there's no problem in this case, what do you think about it?

aguerra: No problem, go ahead.

pkg/server/exec/handlers.go
aguerra: Excelent.

PR #404, opened 2018-02-03
Remove unnecessary interfaces of k8s pkg

drgarcia1986:
Golang is smart enough to infer the interfaces on business packages.

### DIFF
--- a/pkg/server/cmd/run.go
+++ b/pkg/server/cmd/run.go
@@ -53,7 +53,7 @@ func runServer(cmd *cobra.Command, args []string) {
 		log.WithError(err).Fatal("failed to configure storage")
 	}
 
-	k8s, err := getK8s()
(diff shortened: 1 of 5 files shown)


yagonobre:
Maybe an entry in the changelog?

drgarcia1986:
@yagonobre IDK, that diff doesn't change the behavior of the application, but I'm not against the changelog, @aguerra  what do you think about it?

aguerra:
@drgarcia1986 I like the rule no change in behavior, no changelog

drgarcia1986:
@aguerra about the gold rule `three letter variable names`, LGTM 👍 but, `k8s` is a pkg name on Teresa, is there another suggestion?

drgarcia1986:
@aguerra what about `kc`? (ok, I know, is only 2 letters 😝  )

PR #407, opened 2018-02-05
Bc refactor docker

correiabrux:
Refactoring Dockerfile for image resize.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,18 @@
-FROM golang:1.8
+FROM golang:1.8 AS builder
 
-RUN mkdir -p /go/src/github.com/luizalabs/teresa
 WORKDIR /go/src/github.com/luizalabs/teresa
 COPY . /go/src/github.com/luizalabs/teresa
 
 RUN make build-server
+
+FROM debian:9-slim
+RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,11 +1,18 @@
...
 COPY . /go/src/github.com/luizalabs/teresa
 
 RUN make build-server
 
+FROM debian:9-slim
+RUN apt-get update && \
+apt-get install ca-certificates -y &&\
+rm -rf /var/lib/apt/lists/*
+
+WORKDIR /go/src/github.com/luizalabs/teresa
```
yagonobre: Maybe use `/app` as workdir

drgarcia1986: I prefer to follow the go pattern

yagonobre: IMO it's valid only in builder step

drgarcia1986: Gotcha 👍

drgarcia1986:
@correiabrux could you update the CHANGELOG ?

Dockerfile
aguerra: Take a look if the cache dir is /var/lib/apt/lists or /var/cache/apt/archives (maybe both).

PR #409, opened 2018-02-06
Upgrade Golang version of server to 1.9

drgarcia1986:


drgarcia1986:
@aguerra Not necessarily so 😢

PR #410, opened 2018-02-06
Move delete pod on a context cancelation from pkg deploy to pkg exec

drgarcia1986:


pkg/server/exec/exec.go
```diff
@@ -4,21 +4,24 @@ import (
 	"fmt"
 	"io"
 
+	context "golang.org/x/net/context"
```
aguerra: We need an import standard...

pkg/server/exec/exec_test.go
```diff
@@ -91,10 +101,25 @@ func TestRunCommandBySpecNoZeroExitCode(t *testing.T) {
...
 		t.Errorf("expected ErrNonZeroExitCode, got %v", err)
 	}
 }
+
+func TestOpsRunCommandBySpecContextCancelation(t *testing.T) {
+	k8sOps := &fakeK8sOperations{podRunDelay: 10}
+	ops := NewOperations(app.NewFakeOperations(), k8sOps, storage.NewFake(), &Defaults{})
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
```
aguerra: Nice trick calling cancel asap

PR #412, opened 2018-02-07
Add support to deploy local and remote files

aguerra:
For now only http and https are supported.


pkg/client/tar/tar.go
```diff
@@ -92,3 +92,72 @@ func CreateTemp(dir, prefix string, ignorePatterns []string) (string, error) {
...
 	return tmp.Name(), nil
 }
+
+func ExtractToTemp(filename string) (string, error) {
+	gr, rc, err := newReadClosers(filename)
+	if err != nil {
+		return "", err
+	}
+	defer rc.Close()
+	defer gr.Close()
```
drgarcia1986: 💅 
```go
defer func() {
    rc.Close()
    gr.Close()
}()
```

pkg/client/cmd/deploy.go
```diff
@@ -161,13 +166,34 @@ func deployApp(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		client.PrintErrorAndExit("Error accessing path: %v", err)
+	}
+	if finfo.Mode().IsRegular() {
+		path, err = tar.ExtractToTemp(path)
+		if err != nil {
+			client.PrintErrorAndExit("Error extracting archive: %v", err)
+		}
+		defer os.RemoveAll(path)
+	}
```
drgarcia1986: What do you think about to break this code snippet in a small function?

aguerra: Yeap, deploy is already too big

pkg/client/tar/tar.go
aguerra: I guess it shoud be
```go
defer func() {
    gr.Close()
    rc.Close()
}()
```

drgarcia1986: you're right

PR #415, opened 2018-02-09
Add support to filter by pod name to the logs command

aguerra:


pkg/server/app/app_test.go
```diff
@@ -401,7 +401,7 @@ func TestAppOperationsLogs(t *testing.T) {
 		Users: []database.User{*user},
 	}
 
-	rc, err := ops.Logs(user, app.Name, 10, false)
+	rc, err := ops.Logs(user, app.Name, &LogOptions{Lines: 10, Follow: false})
```
yagonobre: maybe create a `opts` var too, like:
```
opts := &LogOptions{Lines: 10, Follow: false}
rc, err := ops.Logs(user, app.Name, opts)
```

PR #421, opened 2018-02-16
Change validation of error on namespace create

drgarcia1986:
To return a friendly error message when an invalid app name is used
in app create method

### DIFF
--- a/pkg/server/app/app_test.go
+++ b/pkg/server/app/app_test.go
@@ -34,6 +34,7 @@ type errK8sOperations struct {
 	SetNamespaceLabelsErr      error
 	DeletePodErr               error
 	NegateIsNotFound           bool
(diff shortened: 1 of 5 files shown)


pkg/server/app/app_test.go
```diff
@@ -201,15 +206,16 @@ func (e *errK8sOperations) Limits(namespace, name string) (*Limits, error) {
...
+func (*errK8sOperations) IsInvalid(err error) bool {
 	return true
 }
 
 func (e *errK8sOperations) IsNotFound(err error) bool {
-	if e.NegateIsNotFound {
-		return false
-	}
-	return true
+	return !e.NegateIsNotFound
```
aguerra: 🥇

PR #423, opened 2018-02-21
Refactor specs to be more in line with k8s concepts

aguerra:
A Pod has one container (for now) and volumes. A container has volume
mounts.


aguerra:
Nope, the magic of embedding.

PR #425, opened 2018-02-23
Make the slugrunner use an init container to download the app code

aguerra:
So it doesn't mount the storage keys anymore.

### DIFF
--- a/pkg/server/exec/exec.go
+++ b/pkg/server/exec/exec.go
@@ -27,6 +27,7 @@ type K8sOperations interface {
 
 type Defaults struct {
 	RunnerImage  string
+	StoreImage   string
 	LimitsCPU    string
 	LimitsMemory string
 }
@@ -56,10 +57,14 @@ func (ops *ExecOperations) RunCommand(ctx context.Context, user *database.User, 
 		return nil, errChan
 	}
 
+	imgs := &spec.SlugImages{
+		Runner: ops.defaults.RunnerImage,
+		Store:  ops.defaults.StoreImage,
+	}
 	podSpec := spec.NewRunner(
 		fmt.Sprintf("exec-command-%s-%s", appName, uid.New()),
 		currentSlug,
-		ops.defaults.RunnerImage,
+		imgs,
 		a,
 		ops.fs,
 		&spec.ContainerLimits{
--- a/pkg/server/k8s/converters_test.go
+++ b/pkg/server/k8s/converters_test.go
@@ -87,7 +87,7 @@ func TestPodSpecToK8sContainer(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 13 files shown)


pkg/server/spec/pod_test.go
```diff
@@ -86,6 +99,13 @@ func TestNewBuilder(t *testing.T) {
 	if ps.ContainerLimits.Memory != expectedContainerLimits.Memory {
 		t.Errorf("expected %s, got %s", expectedContainerLimits.Memory, ps.ContainerLimits.Memory)
 	}
+
+	if len(ps.VolumeMounts) != 1 {
+		t.Errorf("expected %d, got %d", 1, len(ps.VolumeMounts))
```
drgarcia1986: 💅 `t.Errorf("expected 1, got %d", len(ps.VolumeMounts))`

pkg/server/k8s/converters_test.go
```diff
@@ -103,6 +103,25 @@ func TestPodSpecVolumesToK8sVolumes(t *testing.T) {
 	}
 }
 
+func TestPodSpecEmptyDirVolumeToK8s(t *testing.T) {
+	vols := []*spec.Volume{
+		{Name: "Vol-Test", EmptyDir: true},
+	}
+	k8sVols := podSpecVolumesToK8sVolumes(vols)
+
+	for idx, vol := range vols {
```
yagonobre: Maybe remove this for?

pkg/server/exec/exec.go
```diff
@@ -56,10 +57,14 @@ func (ops *ExecOperations) RunCommand(ctx context.Context, user *database.User,
 		return nil, errChan
 	}
 
+	imgs := &spec.SlugImages{
+		Runner: ops.defaults.RunnerImage,
+		Store:  ops.defaults.StoreImage,
+	}
```
drgarcia1986: I'm a bit curious, you created this struct to avoid another argument on `spec` functions?

aguerra: Yep, the  argument list is gigantic.

pkg/server/k8s/converters_test.go
aguerra: In fact, the function is slice based (the slices are not used for test cases) so I guess is easier to use a loop for testing.

yagonobre: Make sense!

aguerra:
@yagonobre @drgarcia1986 Updated. Also put SLUG_DIR as protected env var.

pkg/server/exec/exec.go
drgarcia1986: hm, ok 👍 for a next refactor maybe we could create a struct with all args, something like `options{}`

PR #432, opened 2018-03-05
Add support for internal apps

aguerra:
Internal apps don't have an external endpoint being only visible
inside the cluster.


README.md
```diff
@@ -115,6 +115,12 @@ You also have to adjust the `Procfile` to have a corresponding `worker` key.
 There's nothing special with the name `worker`, it can be anything different
 from `web` with a matching `Procfile` line.
 
+**Q: How to create an internal app (without a external endpoint) ?**
+
+    $ teresa app create <app-name> --team <team-name> --internal
+
+The app is only visible inside the cluster.
```
drgarcia1986: What about an observation about `--process-type` flag? something like "the flag `--internal` only works with `--process-type=web`, I don't know how to describe this, but I think its something relevant.

aguerra: Makes sense.

PR #436, opened 2018-03-05
Add nginx support

yagonobre:


pkg/server/k8s/converters.go
```diff
@@ -84,10 +84,13 @@ func podSpecVolumesToK8sVolumes(vols []*spec.Volume) []k8sv1.Volume {
 		vol := k8sv1.Volume{Name: v.Name}
 		if v.EmptyDir {
 			vol.EmptyDir = &k8sv1.EmptyDirVolumeSource{}
-		} else {
+		} else if v.SecretName != "" {
 			vol.Secret = &k8sv1.SecretVolumeSource{
 				SecretName: v.SecretName,
 			}
+		} else {
```
aguerra: maybe to be more specific and  check for ConfigMapName

pkg/server/spec/container.go
```diff
@@ -5,6 +5,33 @@ import (
...
+
+type VolumeMounts struct {
+	Name      string
+	MountPath string
+	ReadOnly  bool
+}
+
+type Port struct {
+	Name          string
+	ContainerPort int32
```
aguerra: Any specific reason for using named ports or it's just a choice?

yagonobre: I guess that if we have more than 1 one port, this need to be named.

pkg/server/k8s/converters.go
yagonobre: Make sense

pkg/server/spec/pod.go
```diff
@@ -59,7 +59,7 @@ func NewPod(name, nginxImage, image string, a *app.App, envVars map[string]strin
 	ps := &Pod{
 		Name:       name,
 		Namespace:  a.Name,
-		Containers: newPodContainers(image, nginxImage, envVars, hasNginx),
+		Containers: newPodContainers(name, image, nginxImage, envVars, hasNginx),
```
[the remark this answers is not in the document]
aguerra: Ok, I got it, you are using `req.Name` as the container  in `LogOptions`.

pkg/server/spec/pod.go
```diff
@@ -25,26 +25,44 @@ type Pod struct {
...
+	if hasNginx {
+		volumes = append(volumes, &Volume{
+			Name:          "nginx-conf",
+			ConfigMapName: appName,
+		})
+	}
+	return volumes
+}
+
+func newPodContainers(appImage, nginxImage string, envVars map[string]string, hasNginx bool) []*Container {
```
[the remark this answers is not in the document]
yagonobre: Yes, you're right, but I can't think in another solution too.

pkg/server/spec/pod.go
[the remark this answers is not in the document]
yagonobre: yes :D

pkg/server/spec/container.go
```diff
@@ -36,3 +73,51 @@ func newInitContainers(slugURL, image string, a *app.App, fs storage.Storage) []
...
+	return &Container{
+		Name:  name,
+		Image: image,
+		Env:   envVars,
+		Ports: []Port{{
+			Name:          "app",
+			ContainerPort: int32(port),
+		}},
+	}
+}
```
drgarcia1986: What about a new file for nginx specs? `spec/nginx.go1

pkg/server/spec/pod.go
```diff
@@ -10,70 +10,72 @@ const (
...
+	c := []*Container{
+		newAppContainer(name, appImage, envVars, hasNginx),
+	}
+	if hasNginx {
+		c = append(c, newNginxContainer(nginxImage))
+	}
+	return c
+}
+
+func NewPod(name, nginxImage, image string, a *app.App, envVars map[string]string, fs storage.Storage, hasNginx bool) *Pod {
```
drgarcia1986: Maybe you can change args to get struct `Images` struct and check the flag `hasNginx` with `imgs.NginxImage` field.

pkg/server/spec/container.go
yagonobre: LGTM, but the nginx is just a new container spec

pkg/server/spec/pod.go
yagonobre: Make senses

yagonobre:
@aguerra @drgarcia1986 updated

pkg/server/deploy/config.go
```diff
@@ -60,7 +70,7 @@ func getDeployConfigFilesFromTarBall(tarBall io.ReadSeeker, processType string)
 			return nil, err
 		}
 
-		if hdr.Name != tYamlFilename && hdr.Name != tYamlProcessTypeFileName && hdr.Name != ProcfileFileName {
+		if hdr.Name != tYamlFilename && hdr.Name != tYamlProcessTypeFileName && hdr.Name != ProcfileFileName && hdr.Name != nginxConfFileName {
```
drgarcia1986: What about? :
```go
if !isConfigFile(hdr.Name, tYamlFilename, tYamlProcessTypeFileName, ProcfileFileName, nginxConfFileName) {
    continue
}

// ....
func isConfigFile(name string, confFiles ...string) bool {
    for _, cf := range confFiles {
        if name == cf {
            return true
        }
        return false
}

File: pkg/server/spec/container.go (hunk 5 above)
Comment (drgarcia1986) [reply pkg/server/spec/container.go]: I know... follow your heart ❤️ :shipit:

File: pkg/server/spec/pod.go [hunk 8]
```diff
@@ -10,70 +10,82 @@ const (
...
+		c = append(c, newNginxContainer(nginxImage))
+	}
+	return c
+}
+
+func NewPod(name, nginxImage, image string, a *app.App, envVars map[string]string, fs storage.Storage) *Pod {
+	hasNginx := false
+	if nginxImage != "" {
+		hasNginx = true
+	}
```
drgarcia1986:
```diff
- if nginxImage != "" {
-     hasNginx = true
- }
+ hasNginx = nginxImage != ""
```

pkg/server/spec/pod.go
```diff
@@ -10,70 +10,82 @@ const (
...
+	c := []*Container{
+		newAppContainer(name, appImage, envVars, appPort),
+	}
+	if nginxImage != "" {
+		c = append(c, newNginxContainer(nginxImage))
+	}
+	return c
+}
+
+func NewPod(name, nginxImage, image string, a *app.App, envVars map[string]string, fs storage.Storage) *Pod {
```
drgarcia1986: What about change `nginxImage` + `image` to `Images` struct ?

yagonobre:
@aguerra @drgarcia1986 updated and rebased

PR #438, opened 2018-03-05
Improve helm

yagonobre:


wspi:
I've been testing this branch with Minio and noticed a few things:

# Parameters

the helm parameters `aws.s3.bucket` and `aws.s3.endpoint` are ignored during helm install, I had to modify then later.

# Secrets
The secret `teresa-teresa-user` while using Minio, it's called `teresa-minio-user` it must be changed otherwise the pod won't initiate

# Minio
While setting `useMinio=true` would be easier if it modified the default values of `aws.s3.force_path_style`, `storage_type` and `aws.s3.endpoint`

yagonobre:
Thanks, @wspi! I'm working on this.

helm/README.md
```diff
@@ -52,6 +52,19 @@ The [configuration](#configuration) section lists the parameters that can be con
 
 > **Tip**: List all releases using `helm list`
 
+## Install ussing minio as storage backend
```
aguerra: Typo: using

yagonobre:
@wspi now, if you set `useMinio` this will config everything about minio

PR #439, opened 2018-03-06
Fix cronjob creation when the schedule is missing

aguerra:
We return a schedule not found message to the client.


pkg/server/deploy/deploy.go
```diff
@@ -169,6 +169,10 @@ func (ops *DeployOperations) createOrUpdateCronJob(a *app.App, confFiles *Deploy
 		Runner: ops.opts.SlugRunnerImage,
 		Store:  ops.opts.SlugStoreImage,
 	}
+	if confFiles.TeresaYaml == nil || confFiles.TeresaYaml.Cron == nil {
+		errChan <- ErrCronScheduleNotFound
+		return
+	}
```
drgarcia1986: Sorry, I was a kid 👶

PR #444, opened 2018-03-07
Create configMap to nginx deploy, and only deploy nginx if nginx.conf

yagonobre:
👶

### DIFF
--- a/pkg/server/k8s/client.go
+++ b/pkg/server/k8s/client.go
@@ -440,7 +440,7 @@ func (k *Client) Limits(namespace, name string) (*app.Limits, error) {
 	return lim, nil
 }
 
-func (k *Client) createOrUpdateConfigMap(namespace, name string, data map[string]string) error {
+func (k *Client) CreateOrUpdateConfigMap(namespace, name string, data map[string]string) error {
 	kc, err := k.buildClient()
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [NEXT_RELEASE]
 ### Changed
 - Change Method `CreateSecret` to `CreateOrUpdateSecret` on k8s interfaces
+- Only deploy nginx if nginx.conf exist
+- Create configmap on deploy with nginx
```
aguerra: The feature hasn't been released yet so I think you shouldn't add these lines. What do you guys think?

yagonobre: LGTM

drgarcia1986: Agree

pkg/server/k8s/client.go
```diff
@@ -440,7 +440,7 @@ func (k *Client) Limits(namespace, name string) (*app.Limits, error) {
 	return lim, nil
 }
 
-func (k *Client) createOrUpdateConfigMap(namespace, name string, data map[string]string) error {
+func (k *Client) CreateOrUpdateConfigMap(namespace, name string, data map[string]string) error {
```
drgarcia1986: 💅  could you change receiver name to `c`?

PR #445, opened 2018-03-08
Add storage.Storage PodEnvVars on spec of InitContainer

drgarcia1986:
This change fix some issues on get slug with minion as storage backend


drgarcia1986:
@aguerra fixed 👍

PR #450, opened 2018-03-14
Add service command

aguerra:
For now only used to enable ssl.


pkg/server/service/service.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	if err := ops.cops.CreateOrUpdateSSL(appName, cert, sslPort); err != nil {
+		return err
+	}
+	ports := []ServicePort{
+		{Name: defaultPortName, TargetPort: spec.DefaultPort},
+		{Name: defaultSSLPortName, Port: sslPort, TargetPort: spec.DefaultPort},
+	}
+	if only {
+		ports = ports[1:]
```
yagonobre: 👏

PR #468, opened 2018-03-26
Delete configmap if a deploy remove the nginx

yagonobre:


pkg/server/deploy/deploy.go
```diff
@@ -150,6 +152,12 @@ func (ops *DeployOperations) createOrUpdateDeploy(a *app.App, confFiles *DeployC
 			log.WithError(err).Errorf("Creating config to nginx of app %s", a.Name)
 			return
 		}
+	} else {
+		err := ops.k8s.DeleteConfigMap(a.Name, a.Name)
+		if !ops.k8s.IsNotFound(err) {
```
aguerra: It will return a nil error to the channel, right?

yagonobre: It will return the error if it is not `not found`

yagonobre: yes, you're right!

yagonobre: Updated!

drgarcia1986:
Could you update changelog after rebase?

PR #471, opened 2018-03-27
Fix nginx cmd args tests on spec pkg

drgarcia1986:
The func and the test uses maps and we cannot guarantee the order of envs


drgarcia1986:
@aguerra Awesome, I'll try

drgarcia1986:
@aguerra @yagonobre UPDATED

PR #474, opened 2018-03-27
Add a volume in pods with nginx sidecar to share fs

drgarcia1986:
This change allows nginx to serves application static content

(depends on https://github.com/luizalabs/slugrunner/pull/6)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Show CREATED AT instead of AGE in deploy list
 
+### Added
+- Share application directory with nginx sidecar
 
 ## [0.18.0] - 2018-03-27
 ### Added
--- a/pkg/server/spec/deploy.go
+++ b/pkg/server/spec/deploy.go
@@ -87,7 +87,8 @@ func NewDeploy(imgs *Images, description, slugURL string, rhl int, a *app.App, t
 		fs,
 	)
 	ps.Containers[0].Args = []string{"start", a.ProcessType}
-	ps.Containers[0].VolumeMounts = []*VolumeMounts{newSlugVolumeMount()}
+	ps.Containers[0].VolumeMounts = append(ps.Containers[0].VolumeMounts, newSlugVolumeMount())
+
 	ps.InitContainers = newInitContainers(slugURL, imgs.SlugStore, a, fs)
 
 	ds := &Deploy{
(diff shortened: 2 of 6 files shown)


pkg/server/spec/deploy.go
```diff
@@ -87,7 +87,8 @@ func NewDeploy(imgs *Images, description, slugURL string, rhl int, a *app.App, t
 		fs,
 	)
 	ps.Containers[0].Args = []string{"start", a.ProcessType}
-	ps.Containers[0].VolumeMounts = []*VolumeMounts{newSlugVolumeMount()}
+	ps.Containers[0].VolumeMounts = append(ps.Containers[0].VolumeMounts, newSlugVolumeMount())
```
yagonobre: Why change this?

pkg/server/spec/pod.go
```diff
@@ -55,6 +57,12 @@ func newPodContainers(name, nginxImage, appImage string, envVars map[string]stri
 		newAppContainer(name, appImage, envVars, appPort, secrets),
 	}
 	if nginxImage != "" {
+		c[0].VolumeMounts = []*VolumeMounts{
```
yagonobre: Mixed feelings about that, maybe this can be moved to `newAppContainer`.

pkg/server/spec/deploy.go
drgarcia1986: the old way just ignore any volume mounts created before

pkg/server/spec/pod.go
drgarcia1986: `newAppContainer` func doesn't know about nginx right?

yagonobre: `newAppContainer` receive `secondaryPort` as `port` value, but I'm not sure if it's a good approach

drgarcia1986: hm, I'm lost with tons of doubts LOL, @aguerra what do you think about it?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Return error with timeout message when there's no resources for build, release phase and command exec
 
+### Added
+- Share application container file system (directory `/app`) with nginx sidecar
+
```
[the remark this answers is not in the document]
aguerra: I guess  is shorter and to the point

pkg/server/spec/pod.go
aguerra: :shipit:

PR #486, opened 2018-04-11
Add build pkg and some generalizations

drgarcia1986:
This PR is the first part of N about `build` feature.

### DIFF
--- a/pkg/server/build/build.go
+++ b/pkg/server/build/build.go
@@ -1,0 +1,76 @@
+package build
+
+import (
+	"fmt"
+	"io"
+
+	context "golang.org/x/net/context"
+
+	"github.com/luizalabs/teresa/pkg/server/app"
+	"github.com/luizalabs/teresa/pkg/server/exec"
+	"github.com/luizalabs/teresa/pkg/server/spec"
+	"github.com/luizalabs/teresa/pkg/server/storage"
+)
+
+type Operations interface {
+	CreateByOpts(ctx context.Context, opts *CreateOptions) error
+}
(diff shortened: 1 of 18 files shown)


pkg/server/build/build.go
```diff
@@ -0,0 +1,76 @@
...
+type CreateOptions struct {
+	App       *app.App
+	BuildName string
+	SlugIn    string
+	SlugDest  string
+	TarBall   io.ReadSeeker
+	Stream    io.Writer
+}
+
+func (ops BuildOperations) CreateByOpts(ctx context.Context, opt *CreateOptions) error {
```
aguerra: My suggestions regarding build.go: CreateFromOpts, always use `opts` for a pointer to Options and use Operations instead of BuildOperations.

pkg/server/spec/service.go
```diff
@@ -0,0 +1,29 @@
...
+	return &Service{
+		Namespace:  namespace,
+		Name:       name,
+		Type:       sType,
+		TargetPort: tp,
+		Labels:     labels,
+	}
+}
+
+func NewRegularExposeDeploy(appName, sType string) *Service {
```
aguerra: Maybe `NewRegularExposeDeploy` -> `NewDefaultService`

pkg/server/build/build.go
drgarcia1986: LGTM about `opts`.
About `BuildOperations`, BuildOperations is the concrete implementation and Operations is the Interface, what do you think?

PR #487, opened 2018-04-11
Add `app change-team` command

aguerra:


pkg/client/cmd/app.go
```diff
@@ -965,6 +966,44 @@ func appDeletePods(cmd *cobra.Command, args []string) {
...
+	cli := appb.NewAppClient(conn)
+	if err != nil {
+		client.PrintErrorAndExit(client.GetErrorMsg(err))
+	}
+
+	req := &appb.ChangeTeamRequest{AppName: appName, TeamName: teamName}
+	_, err = cli.ChangeTeam(context.Background(), req)
+	if err != nil {
+		client.PrintErrorAndExit(client.GetErrorMsg(err))
+		return
```
drgarcia1986: you don't need that `return` because `PrintErrorAndExit` calls `os.Exit(1)`

PR #491, opened 2018-04-13
Add method build.Create to create (and run) a new build

drgarcia1986:
This PR is the second part of N about build feature.

### DIFF
--- a/pkg/server/build/build.go
+++ b/pkg/server/build/build.go
@@ -7,6 +7,7 @@ import (
 	context "golang.org/x/net/context"
 
 	"github.com/luizalabs/teresa/pkg/server/app"
+	"github.com/luizalabs/teresa/pkg/server/database"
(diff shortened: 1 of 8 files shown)


pkg/server/build/build.go
```diff
@@ -38,7 +53,64 @@ type CreateOptions struct {
...
+	a, err := ops.appOps.Get(appName)
+	if err != nil {
+		errChan <- err
+		return nil, errChan
+	}
+
+	if uhp, err := ops.userHasPermission(a, u); !uhp || err != nil {
+		errChan <- err
+		return nil, errChan
+	}
```
aguerra: Probably I'm missing something, but isn't `appOps.CheckPermAndGet` enough?

drgarcia1986: You're right! I'll fix it.

PR #492, opened 2018-04-16
Add Handler and Client of build feature

drgarcia1986:


pkg/client/cmd/build.go
```diff
@@ -0,0 +1,189 @@
...
+	"io"
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/luizalabs/teresa/pkg/client"
+	"github.com/luizalabs/teresa/pkg/client/connection"
+	"github.com/luizalabs/teresa/pkg/client/tar"
+	bpb "github.com/luizalabs/teresa/pkg/protobuf/build"
+	"github.com/spf13/cobra"
+	context "golang.org/x/net/context"
```
aguerra: It would be nice to split the imports into sections.

Issue #504, opened 2018-05-08
Add a command to remove clusters


aguerra:

drgarcia1986:
fixed on #505

PR #505, opened 2018-05-08
Add rm-cluster command to remove a cluster config

gerep:
The `rm-cluster` command receives the cluster name as the parameter, removes
it from the map variable and rewrites the config file with the remaining
clusters.

Closing #504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [NEXT_RELEASE]
+### Added
+- [client] Add support to remoVe remove cluster from configuration file
+
 ### Changed
 - [server] Update client-go to v6.0.0 and bump minimum k8s version to 1.9
 - [server] Use an explicit selector for the deploy spec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [NEXT_RELEASE]
 ### Added
-- [client] Add support to remoVe remove cluster from configuration file
+- [client] Add support to remove remote cluster from configuration file
 
 ### Changed
 - [server] Update client-go to v6.0.0 and bump minimum k8s version to 1.9
--- a/pkg/client/cmd/config.go
+++ b/pkg/client/cmd/config.go
@@ -61,6 +61,18 @@ eg.:
 	Run: setCluster,
 }
 
+var rmClusterCmd = &cobra.Command{
+	Use:   "rm-cluster name",
+	Short: "removes a cluster entry from the config file",
+	Long: `Remove a cluster entry.
+
+eg.:
+
+  $ teresa config rm-cluster aws_staging
+	`,
+	Run: rmCluster,
+}
+
 var useClusterCmd = &cobra.Command{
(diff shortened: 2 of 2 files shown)


pkg/client/cmd/config.go
```diff
@@ -159,6 +172,38 @@ func setCluster(cmd *cobra.Command, args []string) {
...
+		return
+	}
+
+	c, err := client.ReadConfigFile(cfgFile)
+	if err != nil {
+		c = &client.Config{
+			Clusters:       make(map[string]client.ClusterConfig),
+			CurrentCluster: name,
+		}
+	}
```
aguerra: You guess you don't need to handle this case for rm-cluster, just PrintErrorAndExit.

gerep: @aguerra good point. Something like this instead?

`client.PrintErrorAndExit("Error trying to read config file: %v", err)`

gerep:
@aguerra is there anything else I can do so we can merge this PR?

pkg/client/cmd/config.go
```diff
@@ -159,6 +172,35 @@ func setCluster(cmd *cobra.Command, args []string) {
 	}
 }
 
+// rmCluster removes a server from the config file
+func rmCluster(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
```
drgarcia1986: What about?:
```go
if len(args) == 0 || args[0] == "" {
    cmd.Usage()
    return
}
```

pkg/client/cmd/config.go
```diff
@@ -159,6 +172,35 @@ func setCluster(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		client.PrintErrorAndExit("Error trying to read config file: %v", err)
+	}
+
+	if _, found := c.Clusters[name]; found {
+		delete(c.Clusters, name)
+		if err = client.SaveConfigFile(cfgFile, c); err != nil {
+			client.PrintErrorAndExit("Error trying to save config file: %v", err)
+		}
+		fmt.Printf("Config %s removed", name)
```
drgarcia1986: Could you add a `\n` in end of the text?

aguerra:
Hi @gerep, sorry about the delay (I've disabled github notifications). Now we are back in business so it shouldn't take long to merge it.

pkg/client/cmd/config.go
```diff
@@ -159,6 +172,35 @@ func setCluster(cmd *cobra.Command, args []string) {
...
 
+// rmCluster removes a server from the config file
+func rmCluster(cmd *cobra.Command, args []string) {
+	if len(args) == 0 || args[0] == "" {
+		cmd.Usage()
+		return
+	}
+
+	name := args[0]
+	if name == "" {
```
drgarcia1986: Hm, I think you don't need to check if `args[0]` is empty again, you already did that on line 177.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [NEXT_RELEASE]
+### Added
+- [client] Add support to remoVe remove cluster from configuration file
```
aguerra: remoVe remove -> remove

gerep: Yes, noticed that a bit late, sorry.

PR #511, opened 2018-06-12
Set default backoff limit to 3 on cronjobs

drgarcia1986:
I'll make some tests in a staging environment.


drgarcia1986:
works fine!
@aguerra

PR #512, opened 2018-06-12
Save last user of deploy

drgarcia1986:


pkg/server/deploy/deploy.go
```diff
@@ -113,6 +108,10 @@ func (ops *DeployOperations) Deploy(ctx context.Context, user *database.User, ap
 		if !app.IsCronJob(a.ProcessType) {
 			ops.watchDeploy(appName, deployId, w, errChan)
 		}
+
+		if err := ops.appOps.SaveApp(a, user.Email); err != nil {
```
aguerra: Wouldn't it be better to save the app before watching the deploy?

drgarcia1986: you're right, I'll change

PR #516, opened 2018-06-13
Fix lack of error handling when streaming from readers

aguerra:
Also refactor the goutil package.

### DIFF
--- a/pkg/goutil/channels_test.go
+++ b/pkg/goutil/channels_test.go
@@ -1,26 +1,26 @@
 package goutil
 
 import (
-	"bytes"
+	"strings"
 	"testing"
+	"testing/iotest"
 )
 
-func TestChannelFromReader(t *testing.T) {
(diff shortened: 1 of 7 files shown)


pkg/goutil/channels_test.go
```diff
@@ -1,26 +1,28 @@
...
+	r := iotest.TimeoutReader(strings.NewReader("test"))
+	ch, errCh := LineGenerator(r)
+	<-ch
+	err := <-errCh
 
-		s := <-ch
-		if s != tc.expected {
-			t.Errorf("expected %s, got %s", tc.expected, s)
-		}
+	if err != iotest.ErrTimeout {
```
drgarcia1986: What about ?
```go
if err := <- errCh; err != iotest.ErrTimeout {
    t.Errorf("got %v; want %v", err, iotest.ErrTimeout)
}
```

PR #527, opened 2018-06-27
Add instruction on FAQ about "how to run an app locally"

drgarcia1986:


drgarcia1986:
There are some issues, I'll close to finding the solution.

PR #533, opened 2018-07-01
[Proposal] Improve connection error messages

rafaelgotts:
Since teresa-client user don't need to know how things works, i think that a better error message can help then in day-by-day usage.

Some examples of why i think that can be useful:
- A better message, with explicit instructions to retry **can** save some time.
- A message that have the error, not the cluster configuration, can be helpful to find some real problem.

Here has my proposal changes in use. This was a try to get app info when has no connection to cluster:

**before:**
```
$ teresa app info myapp
Error connecting to server: cluster.address.youalreadyknow:666
```

**after:**
```
$ ./teresa app info myapp
A connection error occour before you command run.
Please, try again.

Error details: context deadline exceeded
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [server] Fix panic when monitoring the rolling update
 - [server] `app info` command output
 
+### Changed
+- [client] Print more friendly error message when connection fails
+
 ## [0.23.0] - 2018-06-21
 ### Added
 - [client] Support to remove remote cluster from configuration file
--- a/pkg/client/errors.go
+++ b/pkg/client/errors.go
@@ -21,3 +21,11 @@ func PrintErrorAndExit(format string, args ...interface{}) {
 	fmt.Fprintln(os.Stderr, color.RedString(format, args...))
 	os.Exit(1)
 }
+
+func PrintConnectionErrorAndExit(err error) {
+	PrintErrorAndExit(
+		"A connection error occoured.\n" +
+		"Please, try again.\n\n" +
+		"Server details: %s", GetErrorMsg(err),
+	)
+}
--- a/pkg/client/errors_test.go
+++ b/pkg/client/errors_test.go
@@ -44,3 +44,19 @@ func TestPrintErrorAndExit(t *testing.T) {
 	}
 	t.Errorf("expected exit status 1, got err %v", err)
 }
+
+func TestPrintConnectionErrorAndExit(t *testing.T) {
+	if os.Getenv("PRINT_CONNECTION_ERROR_AND_EXIT") == "1" {
+		PrintConnectionErrorAndExit(
+			status.Errorf(codes.Unavailable, "Server Unavailable"),
+		)
+		return
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestPrintConnectionErrorAndExit")
+	cmd.Env = append(os.Environ(), "PRINT_CONNECTION_ERROR_AND_EXIT=1")
+	err := cmd.Run()
+	if e, ok := err.(*exec.ExitError); ok && !e.Success() {
+		return
+	}
+	t.Errorf("expected exit status 1, got err %v", err)
+}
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - [server] Fix panic when monitoring the rolling update
 
+### Changed
+- [client] Print more friendly error message when connection fail
+- [client] Return error message instead of cluster config when client connection fail
```
aguerra: connection fails

pkg/client/errors.go
```diff
@@ -21,3 +21,12 @@ func PrintErrorAndExit(format string, args ...interface{}) {
...
 }
+
+func PrintConnectionErrorAndExit(err error) {
+	PrintErrorAndExit(
+		"A connection error occour before you command run.\n" +
+		"Please, try again.\n\n" +
+		"Error details: " +
+		GetErrorMsg(err),
+	)
+}
```
aguerra: I guess this is better :  `A connection error occurred. Please...`
Mixed feelings...for me the error details will only create confusion. @drgarcia1986 What do you think?

pkg/client/errors_test.go
```diff
@@ -44,3 +44,19 @@ func TestPrintErrorAndExit(t *testing.T) {
...
+		return
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestPrintConnectionErrorAndExit")
+	cmd.Env = append(os.Environ(), "PRINT_CONNECTION_ERROR_AND_EXIT=1")
+	err := cmd.Run()
+	if e, ok := err.(*exec.ExitError); ok && !e.Success() {
+		return
+	}
+	t.Errorf("expected exit status 1, got err %v", err)
+}
```
aguerra: For me the the two functions are so close the one doesn't need to test the two, but it's ok.

pkg/client/errors.go
```diff
@@ -21,3 +21,12 @@ func PrintErrorAndExit(format string, args ...interface{}) {
...
 	os.Exit(1)
 }
+
+func PrintConnectionErrorAndExit(err error) {
+	PrintErrorAndExit(
+		"A connection error occour before you command run.\n" +
+		"Please, try again.\n\n" +
+		"Error details: " +
+		GetErrorMsg(err),
+	)
```
drgarcia1986: Use `format` syntax to print the error: `Error details: %s", GetErrorMsg(err)`

pkg/client/errors.go
drgarcia1986: I agree ☝️

pkg/client/errors_test.go
rafaelgotts: The idea of testing was more for coverage points, but it seems that it did not make any difference.

So, i can remove you think is better

pkg/client/errors.go
rafaelgotts: You both think it's better without error details?

rafaelgotts: So, should return the cluster config or just the new message (`A connection error occurred. Please...`)?

drgarcia1986: current cluster config is good to know "where am I?"

pkg/client/errors.go
```diff
@@ -21,3 +21,11 @@ func PrintErrorAndExit(format string, args ...interface{}) {
 	fmt.Fprintln(os.Stderr, color.RedString(format, args...))
 	os.Exit(1)
 }
+
+func PrintConnectionErrorAndExit(err error) {
+	PrintErrorAndExit(
+		"A connection error occoured.\n" +
+		"Please, try again.\n\n" +
+		"Error details: %s", GetErrorMsg(err),
```
drgarcia1986: in this version teresa will print:
```
A connection error occurred.
Please, try again.

Error details: server-url.com.
```
What about?:
```
A connection error occurred.
Please, try again.

Server details: server-url.com.
```

rafaelgotts:
>Could you squash those commits before we merge?
@drgarcia1986 done!

PR #549, opened 2018-07-18
Refactor spec pkg to use DP Builder

drgarcia1986:
This change will facilitate futures features like cloud-sql side-car, volumes, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [server] Database initial migration on mysql 5.7
 - [server] Doesn't log request content on error middleware if the route is `User/Create`
 
+### Changed
+- [server] Big refactor on spec pkg
+
 ## [0.24.1] - 2018-07-11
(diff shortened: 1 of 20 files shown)


drgarcia1986:
@aguerra thank you for your review.

- typos: Already fixed
- s/init/initcontainer/g ✅ 
- _Maybe we will need more than one init container (istio)_, I can change now, but, what about "wait for the problem come"?
- _Using always the last sideCar on the options functions sounds strange to me_, the options funcs of "WithSideCar" method run after the "add" on the sideCar list, so the last sideCar is the sideCar related of the option func.
- _Maybe use WithDefaultLifecycle on the DeployBuilder_: I have some doubts, if we pass the Teresa.yaml to DeployBuilder and use the method WithDefaultLifecycle, maybe we will suffer from misunderstanding like "why Teresa is ignoring my lifecycle custom setting?".
- _Maybe in the tests building the expected struct and using reflect.DeepEqual would be easier_, I'll try, thanks 👍 

Please, let me know what are you thinking about these points.

PR #555, opened 2018-07-30
Add support for cloudsql-proxy sidecar

aguerra:


pkg/server/spec/deploy.go
```diff
@@ -63,6 +64,19 @@ type DeployBuilder struct {
 	d *Deploy
 }
 
+type RawData struct {
+	Fn func(interface{}) error
+}
```
drgarcia1986: 🏆

PR #576, opened 2018-08-21
Add command `app set-vhosts`

aguerra:


pkg/client/cmd/app.go
```diff
@@ -1091,6 +1092,38 @@ func appChangeTeam(cmd *cobra.Command, args []string) {
...
+  $ teresa app set-vhosts myapp myapp.mydomain
+
+  You can also provide more than one vhost at a time:
+
+  $ teresa app set-vhosts myapp myapp.mydomain myapp.anotherdomain`,
+	Run: appSetVHosts,
+}
+
+func appSetVHosts(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
```
drgarcia1986: hm, what about `if len(args) < 2`  because we need at least two args (app name and one vhost)

pkg/server/k8s/client.go
```diff
@@ -625,6 +625,23 @@ func (k *Client) createIngress(namespace, appName string, vHosts []string) error
...
+	if err != nil {
+		return err
+	}
+	old, err := kc.ExtensionsV1beta1().
+		Ingresses(namespace).
+		Get(name, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "get ingress failed")
+	}
+	new := ingressSpec(namespace, name, vHosts)
```
drgarcia1986: what about change `new` to `newSpec` (new is a reserved keyword) ?

pkg/server/k8s/client.go
```diff
@@ -625,6 +625,23 @@ func (k *Client) createIngress(namespace, appName string, vHosts []string) error
...
+	old.Spec.Rules = new.Spec.Rules
+	_, err = kc.ExtensionsV1beta1().Ingresses(namespace).Update(old)
```
drgarcia1986: Is it doesn't work if you only pass the new spec to `Update` method?

pkg/client/cmd/app.go
aguerra: It makes  sense.

pkg/server/k8s/client.go
aguerra: :1st_place_medal:

pkg/server/k8s/client.go
aguerra: It works but overwrites the tls section that might be configured manually.

aguerra: @drgarcia1986

drgarcia1986: gotcha 👍

PR #577, opened 2018-08-21
Add Ubuntu installer

allandieguez:
# Context:

There are no installation instructions for the Ubuntu OS, making it difficult to starting with Teresa right away.

# What was Done:

- Add `install` folder w/ an Ubuntu installation script;
- Update the `README` file to expose the script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,3 +64,6 @@ Run the following in your command-line:
 $ brew tap luizalabs/teresa-cli
 $ brew install teresa
 ```
+
+## Installing Teresa Client:
+ - UBUNTU: execute [this script](./install/install-ubuntu.sh) in the local machine.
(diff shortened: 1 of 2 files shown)


drgarcia1986:
Is it really necessary?
IMO it's very easy to download the binary from [releases](https://github.com/luizalabs/teresa/releases) page.
I'd like to hear the public opinion.

yagonobre:
To me this sounds like over engineering, a note on Readme should be enough, also in `DEPENDENCIES CHECK` you check if `curl` is installed but you use `wget` to download the binary

allandieguez:
Seems fair to me, I'll close this branch and open another PR with a Readme note. Thanks for your input!

PR #581, opened 2018-08-27
Add validation for env var and secret names

aguerra:


pkg/server/validation/envvars.go
```diff
@@ -0,0 +1,29 @@
+package validation
+
+import (
+	"regexp"
+)
+
+var (
+	envVarNameRegexp = regexp.MustCompile("^[-._a-zA-Z][-._a-zA-Z0-9]*$")
```
drgarcia1986: what about?
`^[-._a-zA-Z][-._\w]*$`

aguerra: Sure, I copied from k8s source code :see_no_evil:

drgarcia1986: 😂 it's a minor change, no problem for me

PR #596, opened 2019-04-25
If applied this commit will add one click way to delete a app passing…

alansilva-luizalabs:
Add one click way to delete a app passing --confirm flag;

### DIFF
--- a/pkg/client/cmd/app.go
+++ b/pkg/client/cmd/app.go
@@ -265,21 +265,30 @@ func appDel(cmd *cobra.Command, args []string) {
 	}
 
 	fmt.Print()
-	inputMsg := fmt.Sprintf(
-		"Are you sure you want to delete %s on %s? (yes/NO) ",
-		color.CyanString(name),
-		color.YellowString(currentClusterName),
-	)
-	s, _ := client.GetInput(inputMsg)
-	if s != "yes" {
-		fmt.Println("Delete process aborted!")
-		return
+
+	noinput, err := cmd.Flags().GetBool("no-input")
+	if err != nil {
+		client.PrintErrorAndExit("Invalid description parameter")
 	}
 
-	resp, _ := client.GetInput("Please re type the app name: ")
-	if resp != name {
-		fmt.Println("Delete process aborted!")
-		return
+	if !noinput {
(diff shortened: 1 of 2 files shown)


pkg/client/cmd/app.go
```diff
@@ -265,22 +265,34 @@ func appDel(cmd *cobra.Command, args []string) {
...
-		color.YellowString(currentClusterName),
-	)
-	s, _ := client.GetInput(inputMsg)
-	if s != "yes" {
-		fmt.Println("Delete process aborted!")
-		return
+
+	confirm, err := cmd.Flags().GetString("confirm")
+	if err != nil {
+		client.PrintErrorAndExit("Invalid app name")
```
yagonobre: `Invalid confirm parameter`

pkg/client/cmd/app.go
```diff
@@ -265,22 +265,34 @@ func appDel(cmd *cobra.Command, args []string) {
...
+			fmt.Println("Delete process aborted!")
+			return
+		}
+
+		resp, _ := client.GetInput("Please re type the app name: ")
+		if resp != name {
+			fmt.Println("Delete process aborted!")
+			return
+		}
+	} else if confirm != name {
```
yagonobre: You can remove this else.

pkg/client/cmd/app.go
yagonobre: Also should it be a boolean?

vinicyusmacedo: From what I understand, this `--confirm` flag will actually receive the app name. 

I'm not sure if passing the app name is a good approach. For example, if you use the `--force` flag on git, you're probably aware that the action you're about to do might be destructive.

aguerra: I think `--force` is much clearer (maybe `--silent`).

aguerra:
Watch out for uploading binaries and if possible don't use @alansilva-luizalabs, use your personal github account as you might be removed from Contributors in the future (true story bro).

alansilva-luizalabs:
Hi @yagonobre  @aguerra @vinicyusmacedo. I will open other Pull Request with all suggestions. Thanks.

PR #597, opened 2019-04-26
If applied this commit will add one click way to delete a app passing…

alanwikid:
Add --no-input silence flag to delete app;

### DIFF
--- a/pkg/client/cmd/app.go
+++ b/pkg/client/cmd/app.go
@@ -265,21 +265,30 @@ func appDel(cmd *cobra.Command, args []string) {
 	}
 
 	fmt.Print()
-	inputMsg := fmt.Sprintf(
-		"Are you sure you want to delete %s on %s? (yes/NO) ",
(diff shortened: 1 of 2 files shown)


pkg/client/cmd/app.go
```diff
@@ -265,21 +265,30 @@ func appDel(cmd *cobra.Command, args []string) {
...
+		client.PrintErrorAndExit("Invalid no-input parameter")
 	}
 
-	resp, _ := client.GetInput("Please re type the app name: ")
-	if resp != name {
-		fmt.Println("Delete process aborted!")
-		return
+	if !noinput {
+		inputMsg := fmt.Sprintf(
+			"Are you sure you want to delete%s on %s? (yes/NO) ",
```
drgarcia1986:
```diff
- to delete%s
+ to delete %s
```

alanwikid: Thanks @drgarcia1986

PR #599, opened 2019-05-13
Added support for GCE static IP and SSL

vinicyusmacedo:
On GKE, we can't properly add a SSL certificate to the k8s-created Load Balancer. To be able to set a SSL certificate to an app, we have to manually:

- Reserve an external static IP using gcloud;
- Create an ingress and add the annotations `ingress.gce.kubernetes.io/pre-shared-cert` for the certifcate and `kubernetes.io/ingress.global-static-ip-name` for the static IP;
- Change the service type to NodePort.

This PR adds:

- GCE cloud provider implementation;
- SSL support for GCE;
- CreateOrUpdateStaticIp - a function that adds the `kubernetes.io/ingress.global-static-ip-name` to an ingress; 
- The command `set-static-ip` on `service`, to add the static IP annotation to an ingress (the app has to be created using `--vhost` or `--reserve-static-ip` before using this command, otherwise it'll have no effect);
- The `--reserve-static-ip` flag, which exposes an app using Ingress and sets the static IP annotation.


pkg/protobuf/app/app.proto
```diff
@@ -43,8 +43,9 @@ message CreateRequest {
     Autoscale autoscale = 5;
 
     string virtual_host = 6;
-    bool internal = 7;
-    string protocol = 8;
+    bool reserve_static_ip = 7;
+    bool internal = 8;
+    string protocol = 9;
```
drgarcia1986: hm, maybe is better to maintain the old order and add `reserve_static_ip` as field `9`

pkg/server/app/app.go
```diff
@@ -117,7 +117,7 @@ func (ops *AppOperations) Create(user *database.User, app *App) (Err error) {
 		return auth.ErrPermissionDenied
 	}
 
-	if ops.kops.IngressEnabled() && app.VirtualHost == "" && IsWebApp(app.ProcessType) {
+	if ops.kops.IngressEnabled() && app.VirtualHost == "" && !app.ReserveStaticIp && IsWebApp(app.ProcessType) {
 		return ErrMissingVirtualHost
```
drgarcia1986: What about creating a method called `CheckVirtualHostIsMissing`?

vinicyusmacedo: @drgarcia1986 should I put this method on `app.go`, using `AppOperations` as its receiver?

drgarcia1986: make senses but only on concrete app implementation.

vinicyusmacedo: @drgarcia1986 I didn't quite understand what you mean by the concrete app implementation, but I added the `CheckVirtualHostIsMissing` at https://github.com/luizalabs/teresa/pull/599/commits/e48719479e895e30bb9dfbe85f2f552223f3910a

PR #616, opened 2019-10-19
Added ingress class annotation

vinicyusmacedo:
This is needed, specially on GKE, where you might have two classes of ingress controllers but you are using only one (e.g. nginx ingress). 

This PR might affect existing ingresses though! However, it won't overwrite it to null.

### DIFF
--- a/pkg/server/deploy/deploy.go
+++ b/pkg/server/deploy/deploy.go
@@ -38,7 +38,7 @@ type Operations interface {
 type K8sOperations interface {
 	CreateOrUpdateDeploy(deploySpec *spec.Deploy) error
 	CreateOrUpdateCronJob(cronJobSpec *spec.CronJob) error
-	ExposeDeploy(namespace, name, svcType, portName string, vHosts []string, reserveStaticIp bool, w io.Writer) error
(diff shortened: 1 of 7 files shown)


pkg/server/deploy/deploy.go
```diff
@@ -38,7 +38,7 @@ type Operations interface {
 type K8sOperations interface {
 	CreateOrUpdateDeploy(deploySpec *spec.Deploy) error
 	CreateOrUpdateCronJob(cronJobSpec *spec.CronJob) error
-	ExposeDeploy(namespace, name, svcType, portName string, vHosts []string, reserveStaticIp bool, w io.Writer) error
+	ExposeDeploy(namespace, name, svcType, portName string, vHosts []string, reserveStaticIp bool, ingressClass string, w io.Writer) error
```
drgarcia1986: What about creating a new struct to using as input parameters of this method?

vinicyusmacedo: Good idea.

vinicyusmacedo: Oops, looking at it, I would have to change it in a lot of places. I prefer to open another PR for that.

PR #618, opened 2019-12-23
Expand lifecycle and probes to sidecars

retpolanne:
If applied, this PR expands the liveness, readiness and lifecycle to sidecars, such as the NGINX sidecar  or the CloudSQL Proxy sidecar. 

However, it can be harmful for deployments using CloudSQL Proxy, as the liveness/readiness probe might not work for it.


drgarcia1986 (approved):
The code is ok, but I have some mixed feelings about it.
I think it will cause tons of mistakes.

PR #620, opened 2020-02-03
ndots 2 hardcoded

fandujar:
https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

Due to bad performance of ndots 5, we decided to set ndots 2 in every deploy and app creation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [next_version] - 2020-02-03
+### Changed
+- Added dnsConfig on deploy
+
 ## [0.31.0] - 2019-12-06
 ### Changed
 - Updated minio version on helm chart
--- a/pkg/server/spec/deploy.go
+++ b/pkg/server/spec/deploy.go
@@ -106,6 +106,20 @@ func (b *DeployBuilder) WithRevisionHistoryLimit(rhl int) *DeployBuilder {
 	return b
 }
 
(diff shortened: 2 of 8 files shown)


pkg/server/spec/deploy.go
```diff
@@ -112,6 +122,16 @@ func (b *DeployBuilder) Build() *Deploy {
 			PreStop: &PreStop{DrainTimeoutSeconds: defaultDrainTimeoutSeconds},
 		}
 	}
+
+	if b.d.DNSConfig == nil {
+		b.d.DNSConfig = &DNSConfig{
+			Options: append([]DNSOptions{}, DNSOptions{
+				Name:  "ndots",
+				Value: "2",
```
drgarcia1986: what about creating a config (with that values as default) to hold this configuration?

drgarcia1986:
Hi, @fandujar thanks!
Could you update the CHANGELOG with this modification?

pkg/server/spec/deploy.go
fandujar: @drgarcia1986 that was my first idea, then I kept hard coded 'cause ndots 2 is the new default value on CoreDNS. I will try to create this config.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [next_version] - 2020-02-03
```
drgarcia1986:
```suggestion
## [NEXT_VERSION]
```

pkg/server/spec/deploy.go
drgarcia1986: You can follow your heart

PR #626, opened 2020-03-23
This PR simplifies ingress

vinicyusmacedo:
When we were developing the Ingress support for GCE, we thought the Ingress spec had to be different. However, after testing the same spec we create for Vhost for a couple of months, we found out that there was no problem in using it.


vinicyusmacedo:
image
I believe it is safe to ignore Codecov for now, as the test that was deleted shouldn't affect the function coverage.

PR #635, opened 2020-06-16
Checks if app already has ingress before trying to reserve a static IP

### LINKED ISSUE #634: Teresa deployment is halted if reserveStaticIP is set and an unmanaged ingress already exists
# Steps to reproduce

1. Deploy Teresa with these vars

```
          - name: TERESA_K8S_INGRESS
            value: "true"
          - name: TERESA_DEPLOY_DEFAULT_SERVICE_TYPE
            value: NodePort
          - name: TERESA_K8S_CHECK_ANOTHER_INGRESS
            value: "true"
          - name: TERESA_DEPLOY_INGRESS_CLASS
            value: gce
```

2. Create an app with --vhost and --reserve-static-ip
3. Create an ingress with a name such as `appname-ingress` on the namespace
4. Try to deploy the app, it will throw this error at the end of the deployment:

```
Operation not implemented for apps using k8s load balancer
```
5. If you set reserveStaticIp to false on the `teresa.io/app` annotation on the namespace, the error won't be thrown.

This bug might have something to do with `ErrNotImplementedOnLoadBalancer `
This Teresa deployment is running on GCE

vinicyusmacedo:
This PR checks if the app has a Teresa managed ingress before trying to reserve a static IP.

Fixes #634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Default nginx side-car image to allow support to lua scripts
 - Ingress spec will be the same for vhost and static IP
+- Checks if app already has ingress before trying to reserve a static IP
 
 ### Fixed
 - Don't create an Ingress for apps without vHosts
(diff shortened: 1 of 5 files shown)


vinicyusmacedo:
@drgarcia1986 done!

PR #637, opened 2020-06-16
Adds rewrite-target annotation, tests for Nginx Annotations…

### LINKED ISSUE #636: 404s when using nginx ingress
This [change](https://github.com/luizalabs/teresa/pull/633/files#diff-a8f74b6b8cc169c5832751c4ab1f5cf7L388) caused a side effect on clusters using Nginx Ingress. 

Apparently, the path `/*` translates to a location that is invalid for Nginx (that doesn't support wildcards), but valid for GLBC (that supports wildcards). 

This can be worked around by adding the annotation `nginx.ingress.kubernetes.io/rewrite-target: /` (refer to https://github.com/kubernetes/ingress-nginx/issues/1120#issue-249825699)

vinicyusmacedo:
This PR adds the `nginx.ingress.kubernetes.io/rewrite-target: /` when ingress class is nginx, adds tests for Nginx Annotations and attempts to create a singleton for building a kubernetes fake client.

Fixes #636

### DIFF
--- a/pkg/server/k8s/client_test.go
+++ b/pkg/server/k8s/client_test.go
@@ -379,12 +379,12 @@ func TestHasAnotherIngress(t *testing.T) {
 	if err := cli.createIngress("test", "ingress2", []string{"xpto.com"}, "nginx"); err != nil {
 		t.Fatal("got unexpected error:", err)
 	}
-	rs, err := cli.HasAnotherIngress("test", "test")
+	hasAnotherIngress, err := cli.HasAnotherIngress("test", "test")
 	if err != nil {
 		t.Fatal("got unexpected error:", err)
 	}
 
-	if rs != true {
-		t.Errorf("got true; want false")
+	if hasAnotherIngress != true {
+		t.Errorf("hasAnotherIngress returned false; want true")
 	}
 }
--- a/pkg/server/k8s/client_test.go
+++ b/pkg/server/k8s/client_test.go
@@ -12,6 +12,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func setupK8sCli(t *testing.T, cli *Client) (func(t *testing.T, cli *Client), error) {
+	return func(t *testing.T, cli *Client) {
+		cli.fake = nil
+	}, nil
+}
+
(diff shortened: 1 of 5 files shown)


pkg/server/k8s/client_test.go
```diff
@@ -374,7 +374,31 @@ func TestHasAnotherIngress(t *testing.T) {
 		t.Fatal("got unexpected error:", err)
 	}
 
-	if rs != false {
+	if rs != true {
 		t.Errorf("got true; want false")
```
drgarcia1986, line 378: Could you explain this change?

pkg/server/k8s/client_test.go
```diff
@@ -374,7 +374,31 @@ func TestHasAnotherIngress(t *testing.T) {
...
+
+	if err := cli.createIngress(namespace, ingName, []string{"xpto.com"}, "nginx"); err != nil {
+		t.Fatal("got unexpected error:", err)
+	}
+
+	ing, err := cli.fake.ExtensionsV1beta1().Ingresses(namespace).Get(ingName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal("got unexpected error:", err)
+	}
+	t.Logf("ingress annotations %v\n", ing.Annotations)
```
drgarcia1986, line 395: Is it necessary?

pkg/server/k8s/client_test.go
vinicyusmacedo: The original test was wrong 😅 . We create another ingress, so HasAnotherIngress should've returned true.

pkg/server/k8s/client_test.go
vinicyusmacedo: I'll keep this here while the PR is a draft, as these tests are still failing. For some reason, the Annotations never return, even if we set them :/

pkg/server/k8s/client_test.go
vinicyusmacedo: It returned false because the k8s state wasn't persisted between function calls.

drgarcia1986: ☝️

vinicyusmacedo: @drgarcia1986 I'll change the error message and the variable names, so it'll become less confusing.

PR #639, opened 2020-07-14
Improve error messages

dutradda:
Add some suggestions about the origin of error

### DIFF
--- a/pkg/server/app/errors.go
+++ b/pkg/server/app/errors.go
@@ -12,8 +12,8 @@ var (
 	ErrInvalidName             = status.Errorf(codes.InvalidArgument, "Invalid App Name")
 	ErrInvalidLimits           = status.Errorf(codes.InvalidArgument, "Invalid Limits")
 	ErrInvalidAutoscale        = status.Errorf(codes.InvalidArgument, "Invalid Autoscale")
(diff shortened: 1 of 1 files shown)


pkg/server/app/errors.go
```diff
@@ -12,8 +12,8 @@ var (
 	ErrInvalidName             = status.Errorf(codes.InvalidArgument, "Invalid App Name")
 	ErrInvalidLimits           = status.Errorf(codes.InvalidArgument, "Invalid Limits")
 	ErrInvalidAutoscale        = status.Errorf(codes.InvalidArgument, "Invalid Autoscale")
-	ErrInvalidEnvVarName       = status.Errorf(codes.InvalidArgument, "Invalid Env Var Name")
-	ErrInvalidSecretName       = status.Errorf(codes.InvalidArgument, "Invalid Secret Name")
+	ErrInvalidEnvVarName       = status.Errorf(codes.InvalidArgument, "Invalid Env Var Name (Could be a secret already set)")
```
vinicyusmacedo, line 15: Maybe use `Invalid Env Var Name, or Env Var already set` instead, since it gives the impression that you can use a secret or env var that is already set (that's just 💅 though).

Also, I think you should swap the message here with the message below
```suggestion
	ErrInvalidEnvVarName       = status.Errorf(codes.InvalidArgument, "Invalid Env Var Name (Could be a env var already set)")
```

PR #643, opened 2020-07-29
install: bash script installation

piiih:
This pull request adds a `install.sh` script that downloads the latest version from releases and install.

The README was also updated with instructions.

:)


teresa
```diff
@@ -0,0 +1,20 @@
+HTTP/1.1 302 Found
```
vinicyusmacedo, line 1: Oops, I think you committed the output. Also, Github gives a 302 when you try to get the binary, you should follow this binary. 

Here's a one liner:

```
curl -L https://github.com/luizalabs/teresa/releases/download/`curl -s -o- https://github.com/luizalabs/teresa/releases/latest | sed 's/.*tag\/\(.*\)\".*/\1/'`/teresa-linux-amd64 -o /usr/local/bin/teresa && chmod +x /usr/local/bin/teresa
```

With that, you can skip checking for wget as well. 

It is preferable to keep your binary on `/usr/local/bin` instead of `/usr/bin`. It will be already set to PATH.

piiih: @vinicyusmacedo Damm, I'll remove this file, I didn't notice :(

I did the changes, could you review again pls :)

vinicyusmacedo: No problems, I'll review it again.

install.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+if [[ $1 == "path" && $2 != "" ]]; then
```
vinicyusmacedo, line 3: From what I understand, if you pass this pass value, it will download the teresa binary to the path, right?

What about making it like this:

```sh
./install.sh /my/path
```

That way, you can just test `if [ "$1" ]`.

install.sh
```diff
@@ -0,0 +1,37 @@
...
+  version=$(curl -s -o- https://github.com/luizalabs/teresa/releases/latest | sed 's/.*tag\/\(.*\)\".*/\1/')
+
+  curl -L -O "https://github.com/luizalabs/teresa/releases/download/$version/teresa-linux-amd64"
+
+  chmod +x teresa-linux-amd64
+
+  mv teresa-linux-amd64 "$2/teresa"
```
vinicyusmacedo, line 10: You may create a variable called `OUTPUT_PATH` that points to `/usr/local/bin/teresa` by default. If a path is provided, change it to `$1/teresa`. This way, you conditional will only need to handle the path, and not the download. This way, you won't have to repeat the same code for downloading it.

vinicyusmacedo:
The suggestions may go like this:

```sh
OUTPUT_PATH=/usr/local/bin/
if [ "$1" ]; then
  OUTPUT_PATH="$1/"
  # Do your set path stuff here
fi
# Download teresa binary here and chmod +x
mv teresa-linux-amd64 "$OUTPUT_PATH/teresa"
```

install.sh
```diff
@@ -0,0 +1,37 @@
...
+  echo "After that, restart your terminal :)"
+elif [[ $1 == "" ]]; then
+  if [[ $EUID -ne 0  ]]; then
+    echo "If you don't specify \`path\` you need to run as root(sudo)" 
+    exit 1
+  fi
+
+  version=$(curl -s -o- https://github.com/luizalabs/teresa/releases/latest | sed 's/.*tag\/\(.*\)\".*/\1/')
+
+  curl -L -O "https://github.com/luizalabs/teresa/releases/download/$version/teresa-linux-amd64"
```
vinicyusmacedo, line 31: You may want to provide a way to be able to change the platform as well. I use Darwin for example, and the other person might use windows. If parsing the args for that becomes complex, try doing something like this: https://gist.github.com/vinicyusmacedo/7da040872120508b4b4bf3efdbd0cd53#file-argparse-sh

install.sh
piiih: It was just to be more expressive, but if you think it's understandable this way, I can change, no problem :)

install.sh
piiih: @vinicyusmacedo The only problem with this way is because of the different messages for different scenarios. I can create a function to abstract the chmod + download, what do you think?

piiih: I did two ways, to be possible to install without `sudo` in a specified path

vinicyusmacedo: I get it @piiih . In that case, you could instead keep `$1` on a variable `PROVIDED_PATH` and, if it is provided, you set `OUTPUT_PATH=$PROVIDED_PATH/teresa` and also use it to check if sudo is needed (if `PROVIDED_PATH` is null, exit 1 and ask for sudo).

